PR #6, opened 2023-10-25
Check for presence of parent isocode during zip normalization in region.valid_zip?

kwiersema:

Fix an issue identified when checking valid_zip? on region `PR`. Since `PR` is a province but its parent `Caribbean` does not have an isocode, zip normalization throws an argument error.

I went with the simplest approach to safety check if the parent has an isocode before sending to `Zip.normalize`.
I added test cases to replicate the issue with `PR` in the `region_validation_test`.

### Testing

Verify that US territories such as `PR` now work with `valid_zip?`

### Checklist


rochlefebvre (approved):
This fixes the error, so it's a step forward.

cejaekl (approved):
This makes things better than before the change, so let's merge it.
In the long run, we need to clean things up so that we can reliably ask `PR` for its "country" and get `US` as an answer.  This might involve changing `.parent` from a single value to an array.

PR #17, opened 2023-10-31
Ensure `Region#has_zip?` returns a boolean for all regions

rochlefebvre:
```ruby
irb(main):001:0> Worldwide.region(code: "US").has_zip?
=> true
irb(main):002:0> Worldwide.region(code: "AW").has_zip?
=> false
irb(main):003:0> Worldwide.region(code: "ZZ").has_zip?
=> nil
```
The predicate `has_zip?` should not respond with `nil` on any region.

Double-negate the statement that returns `nil` for regions having no `show` format.

### Testing
```ruby
irb(main):001:0> Worldwide.region(code: "US").has_zip?
=> true
irb(main):002:0> Worldwide.region(code: "AW").has_zip?
=> false
irb(main):003:0> Worldwide.region(code: "ZZ").has_zip?
=> false
```

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ## [Unreleased]
 
 - Add Region#group and Region#group_name [#15](https://github.com/Shopify/worldwide/pull/15)
-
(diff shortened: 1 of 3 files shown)


cejaekl (approved):
Thank you for catching and fixing this.  ❤️

PR #18, opened 2023-10-31
Region loader updates parent for alternates in world.yml

kwiersema:

Based on feedback in https://github.com/Shopify/worldwide/pull/6, we should update PR to have the US be its parent when the hierarchy is built.

I added a conditional to override the parent of the region when it is defined multiple times in the world.yml hierarchy. This will give preference to the alternates instead of where the region is first defined.

Is this an adequate stop gap or do we need to support multiple parents instead?

This change should fix a bug with Puerto Rico zip validation and have territories behave closer to how we are expecting.

### Testing

Focus testing around territories and other places where alternates are defined to ensure the regions are still behaving as expected.

### Checklist

### DIFF
--- a/lib/worldwide/regions_loader.rb
+++ b/lib/worldwide/regions_loader.rb
@@ -46,6 +46,9 @@ module Worldwide
         @regions << current_region
       end
 
+      if parent.present? && current_region.parent != parent
+        current_region.parent = parent
+      end
       parent&.add_zone(current_region)
       return current_region if children.nil?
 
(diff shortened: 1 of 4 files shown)


lib/worldwide/regions_loader.rb
```diff
@@ -46,6 +46,9 @@ def apply_hierarchy(parent:, code:, children:)
         @regions << current_region
       end
 
+      if !parent.nil? && current_region.parent != parent
```
rochlefebvre, line 49: So the last reference in world.yml wins.  Makes sense to me. Also, nit: `if parent.present? && ...`

devanandersen (approved):

PR #26, opened 2023-11-03
[WIP] Algorithmically handle iso_code zones

devanandersen:

...

...

...

...

### Testing

...

### Checklist


lib/worldwide/region.rb
```diff
@@ -384,6 +385,18 @@ def province_optional?
 
     private
 
+    def iso_to_zone_code(iso_code)
+      return iso_code if iso_code.nil? || iso_code.length < 3
+
+      country_code, iso_code = iso_code.split("-")
+      @iso_code_zones ||= Worldwide::Regions.all.select(&:country?)
```
rochlefebvre, line 392: I worry about the repeated cost of computing the set of all countries having subregions whose legacy code starts with `c.iso_code`. This is an instance method, so we're looking at computing this very often for a long-lived application.

Could we compute and cache this list of qualifying country iso_codes in the class?

kwiersema (approved):

cejaekl (approved):
👍 This works, and looks nice.

🤔 I wonder whether we could be slightly more efficient by removing `Worldwide::Regions.iso_code_zones` and doing this instead?

```patch
diff --git a/lib/worldwide/region.rb b/lib/worldwide/region.rb
index d654cb0..1ab6b0a 100644
--- a/lib/worldwide/region.rb
+++ b/lib/worldwide/region.rb
@@ -397,7 +397,7 @@ module Worldwide
       return iso_code if iso_code.nil? || iso_code.length < 3

       country_code, subdivision_code = iso_code.split("-")
-      return subdivision_code if Worldwide::Regions.iso_code_zones.include?(country_code)
+      return subdivision_code if country_code.casecmp(associated_country.iso_code).zero?
     end
```

PR #28, opened 2023-11-03
added building-number-may-be-in-address2

kunwar45:

In the `BuildingNumberInAddress1` and `BuildingNumberInAddress1OrAddress2` validators in Atlas, we check if the building number is allowed to be in address2. The functionality for that is not entirely in Worldwide. 

Specifically this line in Atlas: https://github.com/Shopify/atlas/blob/4c34ca370b0e63e10f9e6c036bb06a6eaf4726a6/app/models/address_validation/validators/street_address.rb#L63

We still use `CountryDb` there. 

I just added the property as an attribute to region. 

The `.yml` files for the countries that this concerns were already updated; except for `IT.yml` which I updated.

### Checklist

### DIFF
--- a/db/data/regions/IT.yml
+++ b/db/data/regions/IT.yml
@@ -14,6 +14,7 @@ zip_regex: "^(IT?-?)?\\d{5}$"
 zip_example: '00144'
 phone_number_prefix: 39
 building_number_required: true
+building_number_may_be_in_address2: true
 week_start_day: monday
 languages:
   - it
--- a/lib/worldwide/region.rb
+++ b/lib/worldwide/region.rb
@@ -47,6 +47,10 @@ module Worldwide
     # If we require a building number in an address, then this will be true.
     attr_accessor :building_number_required
 
+    # In some countries, an address may have the building number in address2.
+    # If we are allowed to have a building number in address2, then this will be true.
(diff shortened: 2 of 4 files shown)


lib/worldwide/region.rb
```diff
@@ -47,6 +47,10 @@ class Region
     # If we require a building number in an address, then this will be true.
     attr_accessor :building_number_required
 
+    # In some countries, an address may have the building number in address2.
+    # If we are allowd to have a building number in address2, then this will be true.
```
kwiersema, line 51: typo `allowed`

db/data/regions/IT.yml
```diff
@@ -14,6 +14,7 @@ zip_regex: "^(IT?-?)?\\d{5}$"
 zip_example: '00144'
 phone_number_prefix: 39
 building_number_required: true
+building_number_may_be_in_address2: true
```
kwiersema, line 17: is Italy the only country where building number is in address2 or are there others we are missing?

kunwar45: Germany, Spain, and The Netherlands also allow building number in address2. For those countries, the `.yml` files already had this line.

kwiersema (approved):
LGTM!

PR #31, opened 2023-11-03
added autofill_city_enabled property to region

kunwar45:

Add a property to show whether a region has `autofill_city_enabled`.

Added this to `Region` and set default to `false`. The regions that are concerned with this already had the correct field in their respective `.yml` files. 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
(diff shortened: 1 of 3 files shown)


devanandersen (commented):
I think this is already available through something like 

```
Worldwide.region(code: 'SG').field(key: :city).autofill(locale: :en).nil?
```

unless I'm mistaken? Is this introducing something different?

kunwar45:
This makes sense, I can use what @devanandersen pointed out. This change is unnecessary.

PR #32, opened 2023-11-07
Expose alternate names of regions and zones

kwiersema:

Expose the alternate names data we have under a new method `alternate_names`

- Add an attr_accessor on region and modify regions loader to set the data accordingly

...

...

### Testing

Verify unit test coverage.

### Checklist

### DIFF
--- a/test/worldwide/region_test.rb
+++ b/test/worldwide/region_test.rb
@@ -249,5 +249,26 @@ module Worldwide
         assert_equal expected_country, Worldwide.region(code: region_code).associated_country.iso_code
       end
     end
+
+    test "name alternates are returned as expected" do
+      {
+        cz: ["Czechia"],
+        sz: ["Swaziland"],
+        us: ["United States of America"],
+        ca: [],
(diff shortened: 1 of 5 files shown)


test/worldwide/region_test.rb
```diff
@@ -249,5 +249,25 @@ class RegionTest < ActiveSupport::TestCase
         assert_equal expected_country, Worldwide.region(code: region_code).associated_country.iso_code
       end
     end
+
+    test "name alternates are returned as expected" do
+      {
+        cz: ["Czechia"],
+        sz: ["Swaziland"],
+        us: [],
```
rochlefebvre, line 257: FYI I fixed a customer-facing issue for the United States in shopify-i18n several weeks ago: https://github.com/Shopify/shopify-i18n/pull/1647

We should make sure that United States of America resolves to the US region.

cejaekl (approved):

test/worldwide/region_test.rb
kwiersema: addressed

PR #35, opened 2023-11-07
Add australia / canberra timezone and translations

kwiersema:

Rails 7.1 has a new timezone key `Australia / Canberra` that may be used instead of `Australia / Melbourne`. Add this key and translations to each timezone translation file

Added a new `Australia / Canberra` key to each timezone file with it's translation.

### Checklist


rochlefebvre:
I don't have any issue with adding this [deprecated](https://www.timezoneconverter.com/cgi-bin/zoneinfo.tzc?s=default&tz=Australia/Canberra) time zone, per se, but maybe we should mention that it's actually deprecated in the changelog?

kwiersema:
@rochlefebvre @cejaekl - I've updated the changelog to clarify that `Australia/Canberra` is an old deprecated timezone being added.

PR #36, opened 2023-11-07
Improve performance by using hash table for by-code region lookup

cejaekl:

Calls to `Worldwide.region(code: x)` and `Worldwide.region(cldr: x)` had been doing linear searches of all known regions.  Recent experience shows that this has become a problem.  Let's use hash table lookups to speed things up.

Note that we're still doing linear searches when looking up by name. Looking up by name is strongly discouraged in any case. Also, the hash table would be significantly larger, because we'd need to cache all translated versions of each name.  At least for now, I'm going to leave that bit of code as it is.

Before:
```ruby
irb(main):001:0> require 'benchmark'
=> true
irb(main):002:0> puts Benchmark.measure{1000.times{Worldwide.region(code: "US-PR")}}
  0.588466   0.010987   0.599453 (  0.616751)
```

After:
```ruby
irb(main):001:0> require 'benchmark'
=> true
irb(main):002:0> puts Benchmark.measure{1000.times{Worldwide.region(code: "US-PR")}}
  0.127260   0.009308   0.136568 (  0.141930)
```

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Add region name alternates [#32](https://github.com/Shopify/worldwide/pull/32)
 - Cache `Region#parent_name` [#33](https://github.com/Shopify/worldwide/pull/33)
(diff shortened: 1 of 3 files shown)


kwiersema (approved):
LGTM! Thank you for taking this work on

devanandersen (approved):
Thank you

PR #38, opened 2023-11-08
Add `iso_code` to `legacy_code` conversion for zone lookup

devanandersen:

* Following https://github.com/Shopify/worldwide/pull/36, we are successfully handling region lookup for territories and subdivisions in `Regions#region`, but we still need to account for these two code types in `Region#zone`

* Setup for conversion of the search code to legacy code format.

* Problems with this approach?

### Checklist


kwiersema (approved):

lib/worldwide/region.rb
```diff
@@ -484,6 +485,12 @@ def search_prefixes_by_normalized_zip(prefixes:, zip:, allow_partial_zip: false)
       end&.first
     end
 
+    def iso_code_to_legacy_code(iso_code)
+      return iso_code if iso_code.nil? || iso_code.length < 3
+
+      iso_code.split("-").last
+    end
```
cejaekl, line 492: 🤔 Unfortunately, the situation is more complicated than this.  The legacy codes are a bit of a... legacy... mess.

```ruby
irb(main):001:0> Worldwide.region(code: "MX").zone(code: "DF").iso_code
=> "MX-CMX"
irb(main):002:0> Worldwide.region(code: "MX").zone(code: "DF").legacy_code
=> "DF"
```

PR #43, opened 2023-11-10
Add associated_continent attribute to region

Flufd:
Adds an associated_continent attribute to the region class that references the containing continent of the region.

Recursively looks at the parent regions until it finds one that is a continent. This means regions with more than one parent like Puerto Rico will take their first parent into consideration, both of which have a parent continent of "North America". French Polynesia is in Oceania, but one of it's parents is France, which lies in Europe. 

Are there any outliers in the above logic that we might hit with this method, or will the first parent pulled out of the set always be the continent where the region actually physically sits?

This is a new field, so should be no impact.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
-Nil.
+- Add support for `Region#associated_continent` to return the containing continent of the region [#43](https://github.com/Shopify/worldwide/pull/43)
 
 ---
 
--- a/test/worldwide/region_test.rb
+++ b/test/worldwide/region_test.rb
@@ -263,6 +263,31 @@ module Worldwide
       end
     end
 
+    test "associated_continent returns the expected continent" do
+      [
+        # Canada
+        [:ca, "North America"],
+        # Ontario is a province of Canada, which is in North America
+        ["CA-ON", "North America"],
+        # Tokyo is a province of Japan, which is in Asia
+        ["JP-13", "Asia"],
+        # North America itself
+        ["003", "North America"],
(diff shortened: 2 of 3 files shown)


test/worldwide/region_test.rb
```diff
@@ -263,6 +263,27 @@ class RegionTest < ActiveSupport::TestCase
...
+      [
+        # Canada
+        [:ca, "North America"],
+        # Ontario is a province of Canada, which is in North America
+        ["CA-ON", "North America"],
+        # Tokyo is a province of Japan, which is in Asia
+        ["JP-13", "Asia"],
+        # North America itself
+        ["003", "North America"],
+        # Peurto Rico is a territory of the US, which is in North America too
```
rochlefebvre, line 276: nit: Puerto Rico

test/worldwide/region_test.rb
```diff
@@ -263,6 +263,27 @@ class RegionTest < ActiveSupport::TestCase
...
+        ["US-PR", "North America"],
```
rochlefebvre, line 277: This example demonstrates that not every "first parent" leads to a continent.  The Caribbean (029)'s first parent is Latin America (419), and the second is North America (003), a continent 👍

CHANGELOG.md
```diff
@@ -27,7 +27,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
-Nil.
+- Add support for `Region#associated_continent` to return the containing continent of the region
```
rochlefebvre, line 30: Could you update the changelog entry to link this PR?
```markdown
- Add support for `Region#associated_continent` to return the containing continent of the region [#43](https://github.com/Shopify/worldwide/pull/43) by @Flufd
```
Attribution is entirely up to you.

PR #50, opened 2023-12-04
add ES zip prefixes

gabypancu:

...

...

...

...

### Testing

...

### Checklist


db/data/regions/ES.yml
```diff
@@ -27,38 +27,54 @@ zones:
   code: C
   tax: 0.0
   tax_name: VAT
+  postal_code_prefixes:
```
cejaekl, line 30:
```suggestion
  zip_prefixes:
```

PR #57, opened 2023-12-08
Avoid `.present?` and `.blank?` so we don't require Rails

cejaekl:

`.present?` and `.blank?` are not in plain Ruby; they're extensions [provided by Rails](https://apidock.com/rails/Object/blank%3F).  Worldwide (is supposed to) support runinng without Rails, so we need to avoid any Rails-only features.

We already have `Util.blank?` and `Util.present?` that provide drop-in replacements for this functionality; it's just a matter of calling our Util method instead.

### Testing

Before:
```ruby
irb(main):001:0> Worldwide.region(code: "CA").zone(code: "ON").valid_zip?("M1A 1A1")
/Users/cejaekl/src/github.com/Shopify/worldwide/lib/worldwide/zip.rb:93:in `normalize': undefined method `present?' for nil:NilClass (NoMethodError)

        return autofill if autofill.present?
                                   ^^^^^^^^^
	from /Users/cejaekl/src/github.com/Shopify/worldwide/lib/worldwide/region.rb:402:in `valid_zip?'
	from (irb):1:in `<main>'
	from bin/console:14:in `<main>'
```

After:
```ruby
irb(main):001:0> Worldwide.region(code: "CA").zone(code: "ON").valid_zip?("M1A 1A1")
=> true
```

Our current unit testing runs (via `bundle exec rake test`) with Rails, which is why we didn't spot this at test time.  We should probably revisit that... but I think that's a broader question for another day.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ## [Unreleased]
 
 - Allow building number in address2 for DK [#53](https://github.com/Shopify/worldwide/pull/53)
+- Avoid .present? and .blank? so we don't require Rails [#57](https://github.com/Shopify/worldwide/pull/57)
 
 ---
 


rochlefebvre (approved):
The gemspec lists activesupport as one of the three runtime gem dependencies. In a future PR, we will need to remove this requirement and fix what breaks.  This PR is a step in that direction, so thank you!

PR #58, opened 2023-12-08
(bugfix) Zone lookup by name

cejaekl:

This fixes two bugs when looking up a zone by name:
 - Consider the possibility that the zone may not have a name. For example, `ES-EA` (Ceuta and Melilla) is not in the Shopify legacy list of known zones, so `.legacy_name` returns `nil`, and this was preventing zone lookup by name for other zones in Spain, because the search would walk off a `nil` pointer.
 - Search `.name_alternates` entries as well.  This was causing `Worldwide.region(code: "ES").zone(name: "Islas Baleares")` to fail, returing the `ZZ` unknown region.

 - check for `nil` with `&.`
 - add check of `name_alternates` to the search condition

### Checklist


rochlefebvre (approved):
Good catch, @cejaekl !

PR #75, opened 2024-01-29
Relax version constraint on our dependency `i18n`

cejaekl:
We had been requiring `i18n` gem version 1.12.x.
The `i18n` gem is now on version 1.14.1, and we'd like to stay up-to-date.

This upgrade triggers two minor behaviour changes, which require us to adjust our tests.

First, the behaviour if `I18n.t()` has been adjusted.

Before (i18n v1.12):

```ruby
irb(main):001:0> I18n.with_locale(:fr){I18n.t("numbers.latn.formats.decimal.patterns.long.standard.1000", count: 1)}
=> "0 millier"
irb(main):002:0> exit
```

After (i18n v1.14.1):
```ruby
irb(main):001:0> I18n.with_locale(:fr){I18n.t("numbers.latn.formats.decimal.patterns.long.standard.1000", count: 1)}
=> "mille"
```

In other words, `i18n` now respects the special count key `'1':`, preferring it over the key `one:`:

```yaml
decimal:
  patterns:
    default:
      standard: "#,##0.###"
    long:
      standard:
        '1000':
          '1': mille
          many: 0 mille
          one: 0 millier
          other: 0 mille
```

We need to adjust our code very slightly to avoid this edge case, so that we can function with both the old and the new versions of `i18n`.

Second, the phrasing of the exception message when a translation is missing has changed its capitalization from `translation missing` to `Translation missing`.  We work around this by downcasing before checking for the string in our test.

### Checklist

* ~[ ] I have added a CHANGELOG entry for this change (or determined that it isn't needed)~  I am not updating the changelog, because this should not be changing any public-facing functionality of this gem.


lib/worldwide/numbers.rb
```diff
@@ -242,7 +242,7 @@ def numeric?(character)
     # https://www.unicode.org/reports/tr35/tr35-numbers.html#Compact_Number_Formats
     # > N is divided by the type, after removing the number of zeros in the pattern, less 1.
     def compact_format_scale_factor(matched_type, style:)
-      compact_number_format = Worldwide::Cldr.t("#{COMPACT_DECIMAL_BASE_KEY}.#{style}.standard.#{matched_type}", count: 1, locale: locale)
+      compact_number_format = Worldwide::Cldr.t("#{COMPACT_DECIMAL_BASE_KEY}.#{style}.standard.#{matched_type}", count: 5, locale: locale)
```
cejaekl, line 245: Avoid falling into the difference-in-behaviour of the `'1':` vs `one:` keys by using a different count.

rochlefebvre (approved):
Excellent!

PR #76, opened 2024-01-30
Add XK and AN to continental grouping

Flufd:
Add continent grouping to Kosovo (XK) and Netherlands Antilles (AN)

The `associated_continent` attribute currently returns nil for these. When we translate these regions' continent names in atlas, we fall back to the EN group name. This gives a mixed language experience, even though we have translations for these continents.

I updated the `world.yml` to reflect the geographic positions of these countries, putting AN into the Caribbean, and XK into Southern Europe. This file is meant to reflect the M49 data, but as AN is a deprecated country code, it doesn't appear in the data, and XK is not fully recognised by all of the UN, so I assume that's why it's not there either.

Should this be done another way to avoid overwriting the M49 data?

...

### Testing

...

### Checklist

### DIFF
--- a/db/data/world.yml
+++ b/db/data/world.yml
@@ -78,6 +78,7 @@
       '029': # Caribbean
         - AI # Anguilla
         - AG # Antigua and Barbuda
+        - AN # Netherlands Antilles
         - AW # Aruba
         - BS # Bahamas
         - BB # Barbados
@@ -240,6 +241,7 @@
       - RS # Serbia
       - SI # Slovenia
       - ES # Spain
+      - XK # Kosovo
     '155': # Western Europe
       - AT # Austria
       - BE # Belgium


cejaekl (commented):
I think that adding XK under 039 ("Southern Europe") is, indeed, the right thing to do.  👍 

😕 I'm unsure whether we really want to be adding AN to this list.  In some sense, having AN cause an error by having a null value is a feature, not a bug.  We have run into problems in the past (many, many times...) where code assumed that AN is a "going concern" and got into trouble down the road.  Perhaps @kwiersema or @devanandersen might want to chime in here?

Flufd:
> 'm unsure whether we really want to be adding AN to this list.

Initially I was only looking at fixing XK, and was trying to find other instances where a country didn't have an associated continent, to add in a test to stop regressions. Happy to drop this part and add in a test with AN as a known "country" region without a continent. For my use case this should be fine. What we were seeing was an extra group in our country selector when we were grouping by continent name. As AN does not appear in our country list, it makes little difference.

devanandersen:
> In some sense, having AN cause an error by having a null value is a feature, not a bug. 

Agree - had the Antilles not have been dissolved, I think this PR on its own would make sense. If anything, we should be slowly removing information to be removed for AN, not adding. 

Would you be able to briefly go into your use case?

> Happy to drop this part and add in a test with AN as a known "country" region without a continent.

This I'm hesitant of - AN isn't technically a known "country" anymore - it was dissolved [October 10, 2010](https://en.wikipedia.org/wiki/Netherlands_Antilles).

cejaekl:
> Agree - had the Antilles not have been dissolved, I think this PR on its own would make sense. If anything, we should be slowly removing information for AN, not adding.

The difficulty with removing AN (and the reason why it's still present here) is that we want to be able to process information about historical orders that were placed prior to October 2010, at which time AN was a "going concern".

cejaekl:
> My use case is a country selector in the Shopify admin that allows you to select countries to be added to a market.

IMHO, "Netherlands Antilles" should not be shown in the list, and should not be selectable.  You shouldn't be able to add a country that no longer exists to a market.

We should be removing any "country" for which `.deprecated?` returns true when constructing the list of countries to display:
```ruby
shopify(dev)> Worldwide.region(code: "AN").deprecated?
=> true
```

PR #103, opened 2024-02-08
Add a new human readable time format for less technical users

Max-Leopold:

Currently none of the datetime formats including a time zone are particularly friendly to less technical users. This PR introduces a new datetime format which can be used when the full date including timezone is important and will be shown to less technical users.

The new format will be rendered as `December 26, 2018 at 1:23 pm (GMT+00:00)`. 
I did not use the timezone abbreviations (e.g. `EST`, `PST` etc.) because 
1. They are not available when rendering dates using Ruby's [strftime](https://apidock.com/ruby/DateTime/strftime)
2. The abbreviations aren't known to all users which can lead to confusion. Using GMT + an offset was the most accessible format we could find for less technical users.

The first commit fixes the script to generate the datetime formats. The script was broken as we always used locale symbols, while sometimes requiring the locale to be a string (for example when using [File.join](https://ruby-doc.org/core-2.6.3/File.html#method-c-join) which requires strings).

The second commit adds the new datetime format
- I added a key to the `CLDR_TO_STRFTIME_MAP` to correctly map the `xxx` key from cldr to `%:z` in ruby ([unicode doc](https://unicode.org/reports/tr35/tr35-dates.html#dfst-zone), [ruby strftime doc](https://apidock.com/ruby/DateTime/strftime))
- Added a new pseudo time format using `Worldwide::Cldr.t("calendars.gregorian.formats.time.long.pattern").sub(":mm:ss", ":mm").sub("z", "'(GMT'xxx')'")`.
- Used the new pseudo time format to generate a datetime format using `long_readable_with_gmt_zone_offset` using: 
```ruby
compound_datetime_format(
          [
            "calendars.gregorian.formats.datetime.long.pattern",
            "calendars.gregorian.additional_formats.yMMMd",
            "pseudo.time_long_gmt_offset",
          ],
        )
```

This is the first time of me dealing with time zone formatting and cldr in detail. So please point out everything that doesn't look right.

A new long datetime format which can be used when displaying dates in a long format for less technical users

### Testing

...

### Checklist


devanandersen (approved):
Out of curiosity, is there a reason we went with GMT over UTC?

Max-Leopold:
@trishrempel @devanandersen 

I change it to use `UTC` instead of `GMT` as I think it's the generally accepted way to represent time offsets.

I was also wondering if we should get rid of the parenthesis around the offset. So instead of `(UTC+05:00)` have `UTC+05:00`. This is because in some translations there are already parenthesis around the timezone. E.g. `2018年12月26日 下午1:23 [(UTC+00:00)]`. cc @chelsleathley

PR #109, opened 2024-02-23
new error codes unknown_for_address

DominiqueFlaaa:
Add new strings for the following errors: 
- `zip_unknown_for_address1_and_city`
- `zip_unknown_for_address`
- `city_unknown_for_address`
- `address1_unknown_for_address`
- `address2_unknown_for_address`
- `province_unknown_for_address`

The messages in the above cases are generated when the address validation system cannot find an exact match for the field, but the system is not confident they are invalid. 

Note these strings are not yet being displayed, which is why we've neglected translating them. 

Introduced these codes under the default english yamls. 

Most of the strings are generic, ie `{field} could not be found for address`. This simplifies the error messaging we have today. 

...

No impact yet; will eventually leverage them in atlas_engine when generating the messages in UnmatchedFieldConcernBuilder. 

### Testing

...

### Checklist


rochlefebvre (approved):
The new copy makes sense to me!

kwiersema (approved):
LGTM!

PR #114, opened 2024-02-29
update error strings for address1_unknown_for_address

DominiqueFlaaa:
Introduced the incorrect strings in https://github.com/Shopify/worldwide/pull/109

Used the term `street` instead of `address line 1`. The content team wants to keep it generic as countries may have different address line expectations, ex: for Indian addresses, Street address usually goes in the second line. 

`Street may be incorrect.` --> `Address line 1 may be incorrect.`
`Street may be incorrect.` --> `Address line 2 may be incorrect.`

...

None; these strings are not yet in use. 
...

### Testing

...

### Checklist

### DIFF
--- a/db/data/regions/_default/en.yml
+++ b/db/data/regions/_default/en.yml
@@ -60,7 +60,7 @@ en:
             street_unknown_for_zip: Enter a valid street name for %{zip}.
             building_number_invalid: Building number couldn't be located for %{street},
               %{zip}.
-            unknown_for_address: Street may be incorrect.
+            unknown_for_address: Address line 1 may be incorrect.
           warnings:
             contains_too_many_words: Address line 1 is recommended to have less than %{word_count} words
         address2:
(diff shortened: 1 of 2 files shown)


db/data/regions/_default/en.yml
```diff
@@ -148,7 +148,7 @@ en:
             contains_html_tags: Postal code cannot contain HTML tags.
             unknown_for_province: Enter a valid postal code for %{province}
             unknown_for_street_and_city: Postal code may be incorrect for %{street}, %{city}
-            unknown_for_address: Postal code my be incorrect.
```
DominiqueFlaaa, line 151: typo

db/data/regions/_default/en.yml
```diff
@@ -82,7 +82,7 @@ en:
             street_unknown_for_zip: Enter a valid street name for %{zip}.
             building_number_invalid: Building number couldn't be located for %{street},
               %{zip}.
-            unknown_for_address: Street may be incorrect.
+            unknown_for_address: Address line 1 may be incorrect.
```
rochlefebvre, line 85: Address line 2 may be incorrect.

rochlefebvre (approved):
We incorrectly say "address line 1" in the address2 section. Otherwise the change is good.

DominiqueFlaaa:
thanks for catching that Roch!

PR #119, opened 2024-03-12
Fix broken CLDR rake tasks

sjahu:
The `cldr:data:import` and `cldr:data:patch` rake tasks were broken because this code was moved over from another repo that depended on some Rails convenience methods.

- The first commit in this PR replaces calls to `.present?` and `.blank?` with the corresponding `Util` methods (similar to https://github.com/Shopify/worldwide/pull/57).

- The second commit removes a patch that was raising an exception: `zh-TW` doesn't exist until a few lines below when it's created based on the contents of `zh-Hant`. I could've moved the call to the method that copies that locale up, but instead chose to remove the redundant patch.

- The third commit adds some useful details to the rake task descriptions.

- The fourth commit removes an unnecessary exception from the patch task that causes the task to return a non-zero status if sorting doesn't do anything. No reason for this.

- The fifth commit fixes a bug in the generate task that was creating changes in the generated data files due to generated files not being in the right order. To get files in the right order (right being defined as "the currently committed order") you had to run the rake task twice. This is because the data wasn't being reloaded and resorted everywhere that it should have been.

Note: I'm not 100% sure the committed order actually is the most correct for all files. I don't think all our generated YAML files (let alone the hand-edited ones) are strictly sorted, and they probably should be. That's a task for another day, however.

Now, if you run `cldr:data:import`, `cldr:data:patch`, then `cldr:data:generate`, all three will succeed without creating any changes to the committed data files. This PR was a hell of a lot of work for "nothing" 😉.

### DIFF
--- a/rake/cldr/puller.rb
+++ b/rake/cldr/puller.rb
@@ -45,11 +45,10 @@ module Worldwide
         end
 
         def normalize_components(components)
-          components = components.presence
           components = components.split(",").map(&:strip).map(&:to_sym).sort if components
 
           invalid_selected_components = (components || []) - ::Cldr::Export::Data.components
-          raise ArgumentError, "Requested unknown CLDR components: #{invalid_selected_components}" if invalid_selected_components.present?
+          raise ArgumentError, "Requested unknown CLDR components: #{invalid_selected_components}" if Util.present?(invalid_selected_components)
 
           components ||= DEFAULT_COMPONENTS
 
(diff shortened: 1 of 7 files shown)


rake/cldr/puller.rb
```diff
@@ -45,11 +45,10 @@ def default_branch_sha
         end
 
         def normalize_components(components)
-          components = components.presence
```
sjahu, line 48: This line depends on a rails method, but it's not necessary because `components` comes from `ENV["COMPONENTS"]`, which is `nil` anyways if that env var is unset (not an empty string).

cejaekl (approved):
🏅 Looks good to me.  Thank you very much for fixing this.

trishrempel (approved):

PR #120, opened 2024-03-12
Patch El Salvador department names

### LINKED ISSUE #117: Inconsistent department naming for El Salvador
This is similar to past CLDR data issues for other territories (e.g. #102 for Spain).

Some Salvadorian department names include "Departamento de" and some don't. This is inconsistent and there's no specific reason why some of the departments would have this prefix while others wouldn't (this conclusion is supported by `es` Wikipedia article titles).

I haven't fully audited other languages, but from a quick glance, they're not consistent either and not even consistently inconsistent 😅 .

Let's remove `departamento de`/`department of`/etc from the translations for all languages. We'll need to add these changes to the `patch` rake task.

https://github.com/Shopify/worldwide/blob/081171c6b2093c172d5dbb62ac7029a41cafa80d/data/cldr/locales/es/subdivisions.yml#L4027-L4040

sjahu:
Fixes #117.

I patched the El Salvador department names to remove "department of" in as many languages as I could based on my best judgement and understanding of the names. It's probably overkill to do so many languages, but we've patched issues for only the most popular language in the past then had to come back and fix them in other languages, too, after somebody complains.

I logged an issue about these inconsistent translations with CLDR here: https://unicode-org.atlassian.net/browse/CLDR-17445 .

I skipped a few languages that I can't read, specifically where the word boundaries weren't obvious enough to identify what needed to be changed.


cejaekl (approved):
⭐ Gold star for going above and beyond the call of duty here.  Thank you.

trishrempel (approved):
Looks good!

PR #138, opened 2024-04-11
Handle empty strings zip normalization

devanandersen:

Before:
```ruby
[10] pry(main)> Worldwide::Zip.normalize(country_code: "", zip: nil)
ArgumentError: Must specify exactly one of cldr:, code: or name:. (code: is preferred)
from /Users/devanandersen/.gem/ruby/3.2.1/gems/worldwide-0.10.3/lib/worldwide/regions.rb:28:in `region'
[11] pry(main)>
```

After:
```ruby
irb(main):001:0> Worldwide::Zip.normalize(country_code: "", zip: nil)
=> nil
```

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
-- Nil.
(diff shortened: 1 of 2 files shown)


kwiersema (approved):

cejaekl (approved):
Thank you!

PR #142, opened 2024-05-13
Reposition `zip` in NF (Norfolk Island) `show:` format.

rochlefebvre:
### WHAT are you trying to accomplish with this PR?

The `zip` should come after the "country" name, not before the city. This came up during discussions with AusPost re: our handling of NF.

Before:
```
9 Quality Row
2899 Kingston
Norfolk Island
```

After:
```
9 Quality Row
Kingston
Norfolk Island 2899
```

AusPost actually wants the format to be:
```
9 Quality Row
Kingston
Norfolk Island NSW 2899
Australia
```
... but we're not going all the way to that, because NF is technically _not_ part of AU-NSW, and not all carriers are equipped to handle that worldview.  We follow ISO-3166-2 and treat NF as a separate "country", in part because that enables our software stack to distinguish between shipping to NF and shipping to AU-NSW.

1. Compared all of the defined country region codes against our internal country library gem's own codes: all countries are present in Worldwide.
2. Compared all format.edit and format.show strings between libraries. All edit format patterns match, but NF's show pattern was different.
3. Tracked down the reason behind the discrepancy, and copied over the missing PR change from March 2023 into the internal library (this PR).

### Checklist


gabypancu (approved):
Nice! Will we update this in Worldwide as well?

PR #148, opened 2024-05-14
Add initial additional address field definitions

gabypancu:

Part of https://github.com/Shopify/address/issues/2559

1. Adds a new attribute to the Region class: `additonal_address_fields`
2. Defines `additonal_address_fields` for [all countries in scope](https://docs.google.com/spreadsheets/d/1LQ-WN_LNc4OmpKZUm4SxfQrlKGUW0I8dbPbsPx8JSNg/edit#gid=601847841) for the project, except for Taiwan. Taiwan will be introduced in a follow up PR as it will have a more complex definition.

These additional address field definitions will be used in concatenation (going from an extended address to a standard one) and splitting (going from a standard address to an extended one) methods. See [this prototype branch](https://github.com/Shopify/worldwide/compare/main...additional-address-fields-proto) for an idea of what these methods will look like.

The field definitions are grouped by the standard field they should be concatenated into (e.g. address1 or address2). They include:
- a key (a required identifier - the english name for the field)
- a decorator (an optional char/word to be used in combination with the reserved delimiter when concatenating, e.g. comma). I analyzed 1000 addresses from each country and [noted](https://docs.google.com/document/d/1mFuPUypjAgyPOUcxU9iRooRZkBg8Xb_AeQA9iZw5FxQ/edit#heading=h.n3i0osiau3ba) the most common delimiters used to separate street, house number, and neighborhood. 

For the Philippines and Vietnam, the decorators are actually words. Are there any foreseeable issues with this? When concatenated addresses are split on the non breaking space, these decorator words should be removed from the resulting strings (we don't need to prefix the neighborhood with the word "neighborhood" once it's in the neighborhood field).

Previously we have used the name "house number", but I've changed this to "building number" as it is more accurate.

I have also chosen the American spelling for neighborhood 🤷‍♀️ Let me know if you have any objections.

### Checklist

### DIFF
--- a/db/data/regions/BE.yml
+++ b/db/data/regions/BE.yml
@@ -24,5 +24,11 @@ languages:
 format:
   edit: "{country}_{firstName}{lastName}_{company}_{address1}_{address2}_{zip}{city}_{phone}"
   show: "{firstName} {lastName}_{company}_{address1}_{address2}_{zip} {city}_{country}_{phone}"
+additional_address_fields:
+  address1:
+    - key: streetName
+      required: true
+    - key: streetNumber
+      required: false
 emoji: "\U0001F1E7\U0001F1EA"
 timezone: Europe/Brussels
--- a/db/data/regions/BR.yml
+++ b/db/data/regions/BR.yml
@@ -19,6 +19,19 @@ languages:
 format:
   edit: "{country}_{firstName}{lastName}_{company}_{zip}_{address1}_{address2}_{city}{province}_{phone}"
   show: "{firstName} {lastName}_{company}_{address1}_{address2}_{zip} {city} {province}_{country}_{phone}"
+additional_address_fields:
+  address1:
+    - key: streetName
+      required: true
+    - key: streetNumber
+      decorator: ","
+      required: true
+  address2:
+    - key: address2
+      required: false
+    - key: neighborhood
+      decorator: ","
+      required: false
 emoji: "\U0001F1E7\U0001F1F7"
 localized_data:
 - name: tax_credential_br
--- a/test/worldwide/region_test.rb
+++ b/test/worldwide/region_test.rb
@@ -350,5 +350,27 @@ module Worldwide
         assert_equal expected_value, Worldwide.region(code: region_code).building_number_required
       end
     end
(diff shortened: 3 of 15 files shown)


db/data/regions/BE.yml
```diff
@@ -24,5 +24,9 @@ languages:
 format:
   edit: "{country}_{firstName}{lastName}_{company}_{address1}_{address2}_{zip}{city}_{phone}"
   show: "{firstName} {lastName}_{company}_{address1}_{address2}_{zip} {city}_{country}_{phone}"
+additional_address_fields:
+  address1:
+    - key: street
+    - key: building_number
```
DominiqueFlaaa: Discussed updating `building_number` -> `buildingNumber` and including a key to indicate whether the field is required.

db/data/regions/BR.yml
```diff
@@ -19,6 +19,19 @@ languages:
 format:
   edit: "{country}_{firstName}{lastName}_{company}_{zip}_{address1}_{address2}_{city}{province}_{phone}"
   show: "{firstName} {lastName}_{company}_{address1}_{address2}_{zip} {city} {province}_{country}_{phone}"
+additional_address_fields:
+  address1:
+    - key: streetName
+      required: true
+    - key: streetNumber
+      decorator: ","
+      required: true
```
rochlefebvre, line 28: Requiring street number on BR checkouts is a new requirement. How is this signaled or enforced by servers & frontends?

test/worldwide/region_test.rb
```diff
@@ -350,5 +350,49 @@ class RegionTest < ActiveSupport::TestCase
         assert_equal expected_value, Worldwide.region(code: region_code).building_number_required
       end
     end
+
+    test "additional_address_fields returns values as expected" do
```
rochlefebvre, line 354: I suggest we only test one or two countries instead of enumerating all of them. Otherwise, as the list of supported additional field countries grows, there will be an expectation of updating this list as well.

gabypancu: I was torn about this, but you're right we don't want it to keep growing

PR #164, opened 2024-05-22
Renaming regional locale codes to default ones

lgunn4:

Part of this project: https://github.com/Shopify/store-localization/issues/4214

We want to provide out of the box translations for all the languages we support. 

Renamed locale codes of files ro-ro, bg-bg sk-sk, sl-si, hr-hr to be their default locale code. 

Did I miss anything?

Merchants using these locale codes as an alternate locale will get these translations baked in now. 

### Checklist

* [X] I have added a CHANGELOG entry for this change (or determined that it isn't needed)


sjahu (approved):
Seems legit to me. cc @cejaekl in case any concerns come to mind for you.

trishrempel (approved):
Looks ok to me, added a member of @Shopify/address for code owner approval

PR #165, opened 2024-05-22
Add address field splitting methods

gabypancu:

Resolves https://github.com/Shopify/address/issues/2595

Follow up to https://github.com/Shopify/worldwide/pull/158

This PR introduces methods that parse out additional address fields from standard fields. For example, parsing street name and street number from address1.

I introduced 2 methods to the `Address` class, `split_address1` and `split_address2`, which return a hash of the parsed out additional fields. If no additional fields are defined for the given country, they return an empty hash. When we are not able to parse the string, we assign the entire string to the first defined additional address field (e.g. street name).

Alternatively, I could have introduced getter methods like `street_name` and `street_number`, but as every country has different additional address fields, I felt this would be misleading.

Is this the ideal return value format for these methods?

Have I tested all the edge cases?

### Checklist

### DIFF
--- a/lib/worldwide/address.rb
+++ b/lib/worldwide/address.rb
@@ -142,7 +142,7 @@ module Worldwide
       Util.blank?(errors)
     end
 
-    RESERVED_DELIMITER = " "
+    RESERVED_DELIMITER = "\u00A0"
     def concatenated_address1
       return address1 if address1.present?
 
@@ -153,6 +153,18 @@ module Worldwide
     def concatenated_address2
       additional_fields = region.additional_address_fields["address2"] || []
       concatenate_fields(additional_fields)
+    end
+
+    def split_address1
+      additional_fields = region.additional_address_fields["address1"] || []
+      split_fields_arr = address1&.split(RESERVED_DELIMITER) || []
+      split_fields(additional_fields, split_fields_arr)
+    end
+
+    def split_address2
+      additional_fields = region.additional_address_fields["address2"] || []
+      split_fields_arr = address2&.split(RESERVED_DELIMITER) || []
+      split_fields(additional_fields, split_fields_arr)
     end
 
     private
@@ -462,6 +474,22 @@ module Worldwide
       else
         concatenated_field.blank? ? RESERVED_DELIMITER : "#{decorator}#{RESERVED_DELIMITER}"
       end
+    end
+
+    def split_fields(field_definitions, field_values)
+      return if field_definitions.empty?
+
+      field_definitions.each_with_index.with_object({}) do |(field, i), split_fields|
+        next unless field_values[i].present?
+
+        field_name = field["key"].underscore
+        split_fields[field_name] = strip_decorators(field_definitions[i + 1], field_values[i])
+      end
+    end
+
+    def strip_decorators(next_field_definition, current_field_value)
+      delimiter = next_field_definition&.fetch("decorator", nil) || ""
+      current_field_value&.delete_suffix(delimiter)
     end
   end
 end
--- a/test/worldwide/address_test.rb
+++ b/test/worldwide/address_test.rb
@@ -143,9 +143,108 @@ module Worldwide
     end
 
     test "concatenated_address2 returns address2 concatenated with neighborhood separated by delimiter and decorator" do
-      address = Address.new(address2: "dpto 4", neighborhood: "Centro", country_code: "BR")
+      br_address = Address.new(address2: "dpto 4", neighborhood: "Centro", country_code: "BR")
+      ph_address = Address.new(address2: "apt 4", neighborhood: "294", country_code: "PH")
+      vn_address = Address.new(address2: "apt 4", neighborhood: "Cầu Giấy", country_code: "VN")
 
-      assert_equal "dpto 4, Centro", address.concatenated_address2
+      assert_equal "dpto 4, Centro", br_address.concatenated_address2
+      assert_equal "apt 4 Barangay 294", ph_address.concatenated_address2
+      assert_equal "apt 4, Quận Cầu Giấy", vn_address.concatenated_address2
+    end
+
+    test "split_address1 returns nil when additional address fields are not defined for the country" do
+      address = Address.new(address1: "123 Main Street", country_code: "US")
+
+      assert_nil address.split_address1
+    end
+
+    test "split_address1 returns blank fields when address1 is blank" do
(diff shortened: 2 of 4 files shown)


test/worldwide/address_test.rb
```diff
@@ -143,9 +143,110 @@ class AddressTest < ActiveSupport::TestCase
...
 
-      assert_equal "dpto 4, Centro", address.concatenated_address2
+      assert_equal "dpto 4, Centro", br_address.concatenated_address2
+      assert_equal "apt 4 Barangay 294", ph_address.concatenated_address2
+      assert_equal "apt 4, Quận Cầu Giấy", vn_address.concatenated_address2
+    end
+
+    test "parse_address1 returns empty hash when additional address fields are not defined for the country" do
+      address = Address.new(address1: "123 Main Street", country_code: "US")
+      expected_hash = {}
```
gabypancu, line 157: Is an empty hash desirable? Or do we want to provide an address1 key/value?

lib/worldwide/address.rb
```diff
@@ -463,5 +475,36 @@ def concatenation_delimiter(concatenated_field, decorator, field_index)
...
+        if i < field_values.length
+          next_field = field_definitions[i + 1] unless last_field_in_arr?(field_definitions, i)
+          raw_field_value = field_values[i]
+          field_value = strip_decorators(next_field, raw_field_value)
+        end
+
+        parsed_fields_hash[field_name] = field_value
+      end
+      parsed_fields_hash
+    end
```
mchemanishop, line 495: I feel there may be a way to simplify this for better readability: 
```suggestion
def parse_fields(field_definitions, field_values)
  field_definitions.each_with_index.with_object({}) do |(field, i), parsed_fields|
    field_name = field["key"].underscore

    field_value = if i < field_values.length
                    raw_field_value = field_values[i]
                    next_field = field_definitions[i + 1] unless last_field_in_arr?(field_definitions, i)
                    strip_decorators(next_field, raw_field_value)
                  else
                    ""
                  end

    parsed_fields[field_name] = field_value
  end
end
```
What do you think?

gabypancu: Oh nice, I like it! I forgot about `each_with_object`

mchemanishop (approved):
Nice turnaround!
LGTM (From what I understand what Worldwide of course :p)

test/worldwide/address_test.rb
kennygoff: Thinking that since this address is valid, we should probably return `{ "address1": "..." }` even though there's no actual parsing happening. Based on the above tests for `concatenated_address1` we still return a string falling back to the existing raw address string, if I understood those tests correctly. Makes sense to me that we return valid data back if we have it.

Seems like a safe assumption to say "no additional address fields" implies "use an assumed default structure of one single `address1` field" in this case.

lib/worldwide/address.rb
```diff
@@ -155,6 +155,18 @@ def concatenated_address2
       concatenate_fields(additional_fields)
     end
 
+    def parse_address1
+      additional_fields = region.additional_address_fields["address1"] || []
+      split_fields_arr = address1&.split(RESERVED_DELIMITER) || []
```
rochlefebvre, line 160: Looking at the definition for RESERVED_DELIMITER above, it's not clear in most browsers, IDEs and text editors that we're using a non-breaking space as opposed to a regular space.

image

Can we make that clearer by setting the constant to
```ruby
RESERVED_DELIMITER = "\u00A0" # non-breaking space
```

test/worldwide/address_test.rb
rochlefebvre: I have a preference for returning an empty response, perhaps even a nil.

Assume we have this address object
```ruby
address = Address.new(address1: "123 Cannot Be Split", ...)
```

### returning a hash with `address1:` as the only key
```
split_address1 = address.split_address1 # => { address1: "123 Cannot Be Split" }

if split_address1.key?(:address1)
  # handle as standard address
else
  # handle as extended address
end
```

### returning `nil`
```
split_address1 = address.split_address1 # => nil

if split_address1
  # handle as extended address
else
  # handle as standard address
end
```
The second one is less surprising to my developer brain. The first form requires more work to evaluate and more knowledge of implementation details.

lib/worldwide/address.rb
```diff
@@ -463,5 +475,32 @@ def concatenation_delimiter(concatenated_field, decorator, field_index)
...
+        end
+        parsed_fields[field_name] = field_value
+      end
+    end
+
+    def last_field_in_arr?(arr, i)
+      i == arr.length - 1
+    end
+
+    def strip_decorators(field_definition, raw_field_value)
```
rochlefebvre, line 497: I recommend we rename the first param to `next_field_definition` and the second one to `current_field_value`, to clarify their relation.

Also, I think `String#delete_suffix` should work here.

```ruby
def strip_decorators(next_field_definition, current_field_value)
  delimiter = next_field_definition&.fetch("decorator", nil) || ""
  current_field_value.delete_suffix(delimiter)
end
```

lib/worldwide/address.rb
```diff
@@ -463,5 +475,32 @@ def concatenation_delimiter(concatenated_field, decorator, field_index)
...
+      field_definitions.each_with_index.with_object({}) do |(field, i), parsed_fields|
+        field_name = field["key"].underscore
+        field_value = if i < field_values.length
+          raw_field_value = field_values[i]
+          next_field = field_definitions[i + 1] unless last_field_in_arr?(field_definitions, i)
+          strip_decorators(next_field, raw_field_value)
+        else
+          ""
+        end
+        parsed_fields[field_name] = field_value
```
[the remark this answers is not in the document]
gabypancu: Oh! Silly me, thanks Roch. I had to make a slight tweak so that we still return an empty string for the fields that are not found, but it's way simpler now


```
field_definitions.each_with_index.with_object({}) do |(field, i), parsed_fields|
    field_name = field["key"].underscore
    parsed_fields[field_name] = strip_decorators(field_definitions[i + 1], field_values[i]) || ""
end
```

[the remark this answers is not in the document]
gabypancu: I like that it explicitly shows what fields exist for the country and what was not found, but I suppose a client should already be aware of which fields are relevant, and can infer what fields are blank based on their presence in our response. I'm fine with omitting the blank fields

rochlefebvre (approved):

PR #167, opened 2024-05-24
Create typescript package for additional address fields

kennygoff:

Resolves https://github.com/Shopify/address/issues/2582

This adds a new typescript npm package that will be used to publish additional address fields functionality splitting and concatenating address strings.

Decision details in [NPM Worldwide(Google Doc)](https://docs.google.com/document/d/13C9eV9BxoWmYLqgj_dT3HmHGE6gozCG4bcav6F5iulg/edit#heading=h.39w4dcvfdzp) 

Most of this is initial project boilerplate and scaffolding to publish.

- `package.json` defines basic package info and commands to run
- `rollup.config.mjs` defines the bundle packaging process
- `src/index.ts` & `src/index.test.ts` are temporary just to make sure the tooling is working and don't define the planned API
- New CI workflow for build and release (`.github/workflows/ci-typescript.yml` and `.github/workflows/npm-release.yml`)

Introduces a new package to be published publicly under the Shopify npm org: `@shopify/worldwide`

### Testing

After following the setup and build steps in `lang/typescript/README.md`, you can run `pnpm test` to ensure it correctly runs the jest test suite:

Screenshot 2024-05-24 at 3 43 40 PM

### TODOs

* [x] Add `license` to `package.json`
* [x] Add `description` to `package.json`
* [x] Update `author` in `package.json`
* [x] Setup GH Actions
* [x] Setup initial changelog and version for NPM

### Checklist


lang/typescript/jest.config.js
```diff
@@ -0,0 +1,5 @@
+/** @type {import('ts-jest').JestConfigWithTsJest} */
+module.exports = {
+  preset: 'ts-jest',
+  testEnvironment: 'node',
+};
```
brendanrygus, line 5: Not a blocker, but did we consider [Vitest](https://vitest.dev/) at all?

> Even if your library is not using Vite (for example, if it is built with esbuild or Rollup), Vitest is an interesting option as it gives you a faster run for your unit tests and a jump in DX thanks to the default watch mode using Vite instant Hot Module Reload (HMR). Vitest offers compatibility with most of the Jest API and ecosystem libraries, so in most projects, it should be a drop-in replacement for Jest.
https://vitest.dev/guide/comparisons.html#:~:text=Even%20if%20your,replacement%20for%20Jest.)

We're not using Vite for builds here but Vite uses Rollup under the hood which sets up a nice migration path if you ever need some of those advanced locale loading features mentioned in the tech doc.

[Some internal teams](https://grokt.shopify.io/results?q=vitest.config) have been adopting it for new libraries for the improved performance and ESM support, and it's mostly backwards compatible with Jest if you need to reference existing examples or tools. I just helped the SFN team get it set up for their Remix app recently and it was pretty smooth (aside from needing to run Jest and Vitest in parallel, which you can avoid here).

lang/typescript/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "version": "0.1.0-dev",
+  "repository": "git@github.com:Shopify/worldwide.git",
+  "author": "Shopify Inc.",
+  "license": "SEE LICENSE IN LICENSE.md",
+  "publishConfig": {
+    "access": "restricted"
+  },
+  "main": "dist/index.js",
+  "types": "dist/src/index.d.ts",
+  "packageManager": "yarn@4.2.2",
```
brendanrygus, line 13: One small nuance that might have been missed in the original `yarn` vs `pnpm` comparison: most internal teams using yarn are still on v1 (!). That was the only supported version in shopify-build CI, see this thread for more context: https://github.com/Shopify/shopify-build/issues/4975.

Since this is an open source repository you will likely use Github Actions over Shopify Build and won't face the limitations, so this might be a non-issue. Just noting that if you run into recurring issues with yarn v4+ (like pnp & eslint) there [may not be many other internal examples to work from](https://grokt.shopify.io/results?q=packageManager+file%3Apackage.json) as some of the big apps move to pnpm instead.

lang/typescript/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@shopify/worldwide",
+  "description": "Utilties for parsing and formatting address fields",
```
brendanrygus, line 3:
```suggestion
  "description": "Utilities for parsing and formatting address fields",
```

brendanrygus (commented):
Overall approach looks good to me, and I'm glad to see you're already thinking about the performance considerations for bundling/loading the YAML locales. Just left a couple of comments with more organization context to consider, but not blocking at all.

lang/typescript/jest.config.js
kennygoff: I did not only because I read Jest was used and didn't see much mention of Vitest in vault docs I was looking at and I was trying to go with common defaults within Shopify. Can definitely make the switch though.

DominiqueFlaaa (approved):
tophatted with the following 
```
sudo npm install -g pnpm
cd lang/typescript 
pnpm install
pnpm build
pnpm test
```
lookin good 👍

PR #171, opened 2024-05-28
Add `to_rails` method to convert IANA timezones to supported Rails timezones

alex-amaral:

Add a method to `DeprecatedTimeZoneMapper` to convert IANA timezones to the supported Rails timezones. The logic for the conversion is:

1. If it's a supported Rails timezone, just return the timezone.
2. If it's an alias IANA timezone:
  a. Check if the alias timezone is a Rails timezone. If so, return it.
  b. In case the alias timezone isn't a Rails timezone, get the closest geographic location within the same GMT offset and return this timezone.
3. If it's not a IANA timezone or Rails timezone, just return it.

pr_agent:walkthrough

Are all Rails timezones supported?
Is the conversion logic correct?
Is the logic performant or is there a way to improve the performance?

There is no direct impact since we don't have any change in the `to_supported` method and there is no client of this new method.

### Checklist


lib/worldwide/deprecated_time_zone_mapper.rb
```diff
@@ -8,6 +10,103 @@ class << self
...
+
+        closest_timezone
+      end
+
+      def get_coordinates(timezone)
+        # Get the timezone identifier
+        identifier = timezone.identifier
+
+        # Use the TZInfo::Country class to find the country and its timezones
+        TZInfo::Country.all.each do |country|
```
rochlefebvre, line 97: This is a somewhat expensive scan to perform `1 + neighboring_zones` times on every lookup. In my review comments, I will recommend that we precompute this information, either in a kind of initializer or even as a gem build artifact (a yaml file).

test/worldwide/deprecated_time_zone_mapper_test.rb
```diff
@@ -26,5 +26,432 @@ class TimeZoneTest < ActiveSupport::TestCase
...
+        ["America/Guadeloupe", "America/Puerto_Rico"],
+        ["America/Indianapolis", "America/Indiana/Indianapolis"],
+        ["America/Jujuy", "America/Argentina/Jujuy"],
+        ["America/Knox_IN", "America/Indiana/Knox"],
+        ["America/Kralendijk", "America/Puerto_Rico"],
+        ["America/Louisville", "America/Kentucky/Louisville"],
+        ["America/Lower_Princes", "America/Puerto_Rico"],
+        ["America/Marigot", "America/Puerto_Rico"],
+        ["America/Mendoza", "America/Argentina/Mendoza"],
+        ["America/Montreal", "America/Toronto"],
```
rochlefebvre, line 93: This particular test line does not give much value, as America/Toronto is not supported by Rails (at least, according to the original GH issue that prompted this work).

lib/worldwide/deprecated_time_zone_mapper.rb
```diff
@@ -8,6 +10,103 @@ class << self
...
+        timezone_alias = TIMEZONES_ALIASES.find do |timezone, aliases|
+          return timezone.to_s if aliases.include?(iana_timezone)
+        end
+
+        # If the alias is also a supported Rails timezone, return it
+        return timezone_alias if RAILS_TIMEZONES.key?(timezone_alias&.to_sym)
+
+        timezone = timezone_alias || iana_timezone
+
+        matching_timezones = find_matching_timezones(timezone)
```
cejaekl, line 29: Timezone equivalence is more complicated than just the offset from UTC.  For example, `Africa/Casablanca` and `Africa/Monrovia` are both `UTC+0`, but [Casablanca](https://www.zeitverschiebung.net/en/city/2553604) applies DST part of the year while [Monrovia](https://www.zeitverschiebung.net/en/city/2274895) does not, so we cannot treat them as equivalent timezones.

lib/worldwide/deprecated_time_zone_mapper.rb
```diff
@@ -8,6 +10,103 @@ class << self
...
+
+        return timezone if matching_timezones.empty?
+
+        iana_tz_info = TZInfo::Timezone.get(timezone)
+
+        # Find the closest Rails timezone based on geographical coordinates
+        closest_timezone = find_closest_timezone(iana_tz_info, matching_timezones)
```
cejaekl, line 36: Unfortunately, geographic proximity does not correlate reliably with the timezone that's in effect.  For example, this code ends up mapping `America/North_Dakota/Beulah` (Central time, UTC-6, but with daylight saving observed at UTC-5) to its near-neighbour `America/Regina` (Central time, UTC-6, but does not observe daylight saving, so it's currently an hour different from Beulah as I type this in May 2024):

```ruby
shopify(dev)> Worldwide::DeprecatedTimeZoneMapper.to_rails("America/North_Dakota/Beulah")
=> "America/Regina"
```

I don't think there's a good substitute here for a manual review; fortunately, there are only a few hundred mappings to consider.  On the upside, we can get rid of the dependency on `geokit`, because it doesn't help us.  ;)

cejaekl (commented):
❤️ First and foremost, _thank you_ for caring to make this better.  Implementing time zone mapping is a tricky problem.

I think @rochlefebvre has the right idea here:
> As a quick follow up to this work, maybe we should write a rake task that computes a yaml file that maps every known IANA timezone to its closest Rails timezone.

... but I would go one step further.  We could implement `.to_rails` the same way that the existing `.to_supported` is [implemented](https://github.com/Shopify/worldwide/blob/5bc2466b5323e93cbf38da86ae632228f9389526/lib/worldwide/deprecated_time_zone_mapper.rb#L9):  a hash table lookup.  The number of IANA timezones is small (a few hundred), and they don't change _that_ often.  By precomputing a mapping and then storing it (either in code as a hash table, or as a `.yml` file that gets loaded-and-cached at runtime), we can dramatically speed up the time to map zones in production.

<details>
<summary>In terms of correctness, @rochlefebvre has spotted one problem:</summary>

```ruby
shopify(dev)> Worldwide::DeprecatedTimeZoneMapper.to_rails("America/Montreal")
=> "America/Toronto"
shopify(dev)> Worldwide::TimeZone.all.map(&:name).include?("America/Toronto")
=> false
shopify(dev)> Worldwide::TimeZone.all.map(&:name).include?("America/New_York")
=> true
```

</details>

<details>
<summary>Poking at the code a bit, I can see a few other problem cases:</summary>

```ruby
shopify(dev)> Worldwide::DeprecatedTimeZoneMapper.to_rails("Africa/Accra")
=> "Africa/Abidjan"
shopify(dev)> Worldwide::TimeZone.all.map(&:name).include?("Africa/Abidjan")
=> false
shopify(dev)> Worldwide::TimeZone.all.map(&:name).include?("Africa/Monrovia")
=> true
shopify(dev)> Worldwide::DeprecatedTimeZoneMapper.to_rails("America/Atikokan")
=> "America/Panama"
shopify(dev)> Worldwide::TimeZone.all.map(&:name).include?("America/Panama")
=> false
shopify(dev)> Worldwide::TimeZone.all.map(&:name).include?("America/Lima")
=> true
shopify(dev)> Worldwide::DeprecatedTimeZoneMapper.to_rails("Europe/Nicosia")
=> "Asia/Nicosia"
shopify(dev)> Worldwide::TimeZone.all.map(&:name).include?("Asia/Nicosia")
=> false
shopify(dev)> Worldwide::TimeZone.all.map(&:name).include?("Europe/Athens")
=> true
```

</details>

<details>
<summary>I'd suggest that you add a unit test to verify completeness of your mappings as follows:</summary>

Download the latest IANA tzdata from [here](https://www.iana.org/time-zones), and untar it.

Construct a list of all IANA timezone names:
```bash
cejaekl@MacBook-Pro t3 % grep -o "^Zone\s*[A-Za-z_\/]*" northamerica backward backzone etcetera europe antarctica africa asia australasia southamerica | grep -o "\s[A-Za-z_\/]*$" | grep -o "[A-Za-z_\/]*" >all_zones.txt
cejaekl@MacBook-Pro t3 % grep -o "^Link\s*[A-Za-z_\/]*\s*[A-Za-z_\/]*" northamerica backward backzone etcetera europe antarctica africa asia australasia southamerica | grep -o "\s[A-Za-z_\/]*$" | grep -o "[A-Za-z_\/]*" >all_links.txt
cejaekl@MacBook-Pro t3 % cat all_links.txt all_zones.txt| sort | uniq >combined.txt
cejaekl@MacBook-Pro t3 % wc -l combined.txt
     561 combined.txt
```

You can then read the list of timezones into Ruby at a dev console (or in a unit test):
```ruby
shopify(dev)> combined = File.read("combined.txt").split("\n")
=>
["Africa/Abidjan",
...
```
... and then find out which of those zone names fail to map to a zone known to `Worldwide::TimeZone` like this:
```ruby
shopify(dev)> combined.map{|tz| m = Worldwide::DeprecatedTimeZoneMapper.to_rails(tz); [tz, m, Worldwide::TimeZone.all.map(&:name).include?(m)]}
=>
[["Africa/Abidjan", "Africa/Monrovia", true],
 ["Africa/Accra", "Africa/Abidjan", false],
 ["Africa/Addis_Ababa", "Africa/Nairobi", true],
...
```
</details>

lib/worldwide/deprecated_time_zone_mapper.rb
alex-amaral: Sorry for the silly question, but that's something I'd start to look how to solve. What should be considered in this case? `America/North_Dakota/Beulah` isn't a supported timezone and we should map to something else, but the closest timezone we could use is `America/Regina`. All the others are either to far away (e.g. `America/Bogota` and `America/Lima`) or also outside DST like `America/Regina`.

cejaekl: > Sorry for the silly question, but that's something I'd start to look how to solve. What should be considered in this case? `America/North_Dakota/Beulah` isn't a supported timezone and we should map to something else, but the closest timezone we could use is `America/Regina`. All the others are either to far away (e.g. `America/Bogota` and `America/Lima`) or also outside DST like `America/Regina`.

That's a very good question (not silly at all!).  😄 

I think it helps to consider "why" we want a timezone.  Generally, it's so that we can convert a UTC date/time into a local date/time that the merchant or buyer will understand.  So, for example, if we expect a package to arrive in Toronto by 2024-05-29 at 19:00 UTC, we want to be able to display that in the UI as 2024-05-29 at 5pm.  To that end, for timezones that aren't supported by Rails, we want to find an alternative timezone that will give us the same formatted time result.  In this example, we know that we can use `America/New_York` instead of `America/Toronto`, because the two time zones have the same offsets and DST transition dates in 2024.

(David Olson made a design decision decades ago that tzdb would care about any timezones that presented a different time at any point after the start of the Unix "epoch" in 1970-01-01.  As a result, there are a lot of time zones in the IANA tzdb that only differ in what happened at some point shortly after 1970, and give equivalent answers for any "recent" date.  Continuing with the example above, `America/Toronto` and `America/New_York` have had the same offset and DST transition rules since 1986.) 

In the case of `America/North_Dakota/Beulah`, we need to identify an alternative timezone that
 - is supported by Rails,
 - has the same offsets and DST transition rules in 2024

The tzdb entry for Beulah is:
```
Zone America/North_Dakota/Beulah -6:47:07 - LMT 1883 Nov 18 19:00u
                        -7:00   US      M%sT    2010 Nov  7  2:00
                        -6:00   US      C%sT
```
This means that, from November 2010 to the present, Beulah uses offset `-6:00` with the `US C%sT` rules for DST transitions.

The zone `America/Chicago` has used offset `-6:00` and the `US C%sT` rules for DST transitions since 1967:

```
Zone America/Chicago    -5:50:36 -      LMT     1883 Nov 18 18:00u
                        -6:00   US      C%sT    1920
                        -6:00   Chicago C%sT    1936 Mar  1  2:00
                        -5:00   -       EST     1936 Nov 15  2:00
                        -6:00   Chicago C%sT    1942
                        -6:00   US      C%sT    1946
                        -6:00   Chicago C%sT    1967
                        -6:00   US      C%sT
```

Thus, for any date after 2010, we can safely use `America/Chicago` to localize UTC to the `America/North_Dakota/Beulah` time zone, and can list the former as the `.to_rails()` alias for the latter; it's one of the Rails-supported timezones.

```ruby
shopify(dev)> Worldwide::TimeZone.all.map(&:name).include?("America/Chicago")
=> true
```

To make this concrete:
 - 2024-05-29 19:00 UTC is 2024-05-29 at 4pm in Beulah
 - 2024-05-29 19:00 UTC is also 2024-05-29 at 4pm in Chicago
 - 2024-05-29 19:00 UTC is 2024-05-29 at 3pm in Regina

Therefore, `America/Chicago` is an acceptable alias for `America/North_Dakota/Beulah`, but `America/Regina` is not.

alex-amaral:
Closing in favor of https://github.com/Shopify/worldwide/pull/194

PR #177, opened 2024-05-29
Update additional_address_fields and combined_address_format defn's

DominiqueFlaaa:
These include the changes necessary for presence validation on the additional address fields. 

_Part of https://github.com/Shopify/address/issues/2569_

`Region.rb`
- added `combined_address_format` attribute 
- added the following instance methods
  - `street_name_required?`
  - `street_number_required?`
  - `neighborhood_required?`
- did not add `line_2_required?` as no countries will have this requirement

`Address.rb`
- updated concat + splitting methods to leverage `combined_address_format` 

`regions/*.yml`
- updated `additional_address_fields` definitions to return an array of fields + field requirements
- introduced `combined_address_format` defn to return a hash of the address1 and address2 format 

`TW.yml` 
- added base definitions for fields and combined format 

Added consistency tests to ensure that
1. a region's `additional_address_fields` name values are allowed 
2. a region's `additional_address_fields` are present in the `combined_address_format` definition

Double check the yml definitions. 

Allows callers to check for field requirement. 

`additional_address_fields` will now return an array of fields + field requirements

No impact on concatenation and splitting logic. 

### Testing

...

### Checklist

### DIFF
--- a/lib/worldwide/region.rb
+++ b/lib/worldwide/region.rb
@@ -34,6 +34,7 @@ module Worldwide
       :zip_regex,
       :zip_requirement,
       :additional_address_fields,
+      :combined_address_format,
     ]
 
     # A region may have more than one parent.
@@ -197,8 +198,11 @@ module Worldwide
     # If true, then the province is optional for addresses in this region.
     attr_accessor :province_optional
 
-    # A hash of additional address fields and the rules for concatening them into the standard fields
+    # An array of the additional address fields that are defined for this region
     attr_accessor :additional_address_fields
+
+    # A hash of the rules for concatening the additional address fields into the standard fields
+    attr_accessor :combined_address_format
 
     def initialize(
       alpha_three: nil,
@@ -235,7 +239,8 @@ module Worldwide
       @tax_rate = tax_rate
       @use_zone_code_as_short_name = use_zone_code_as_short_name
 
-      @additional_address_fields = {}
+      @additional_address_fields = []
+      @combined_address_format = {}
       @building_number_required = false
       @building_number_may_be_in_address2 = false
       @currency = nil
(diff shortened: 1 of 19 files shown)


lib/worldwide/region.rb
```diff
@@ -197,9 +198,12 @@ class Region
     # If true, then the province is optional for addresses in this region.
     attr_accessor :province_optional
 
-    # A hash of additional address fields and the rules for concatening them into the standard fields
+    # An array of additional address fields that are required for this region
```
rochlefebvre, line 201: Not all fields are required, at least in the sense that we employ. How about _address fields that are defined_... ?

lib/worldwide/region.rb
```diff
@@ -197,9 +198,12 @@ class Region
...
     attr_accessor :additional_address_fields
 
+    # A hash of the rules for concatening the additional address fields into the standard fields
+    attr_accessor :combined_address_format
+
```
gabypancu, line 206: We should also update the attribute set on line 242
```
@additional_address_fields = {}
```

to

```
@additional_address_fields = []
@combined_address_format = {}
```

lib/worldwide/region.rb
```diff
@@ -441,6 +460,11 @@ def add_zone_to_hash(zone)
       end
     end
 
+    def additional_field_required?(field_name)
+      field = additional_address_fields.find { |f| f["name"] == field_name }
+      field ? field["required"] : false
```
rochlefebvre, line 465: If the field yaml does not include a `required:` key, then `field["required"]` will actually return `nil`. Predicate methods should stick to `true/false`. How about we do:

```ruby
field ? !!field["required"] : false
```

The `!!` will convert all truthy values to `true` and falsy values to `false`, including `nil`.

test/worldwide/region_yml_consistency_test.rb
```diff
@@ -100,6 +100,35 @@ class RegionYmlConsistencyTest < ActiveSupport::TestCase
...
+          field_name = field["name"]
+          present = country.combined_address_format.any? do |_, fields|
+            fields.any? { |f| f["key"] == field_name }
+          end
+
+          assert present, "#{country.iso_code} additional_address_field #{field_name} is not present in combined_address_format"
+        end
+      end
+    end
+
```
gabypancu, line 131: Can we also add a test to make sure all keys in `combined_address_format` are also in `additional_address_fields`?

lib/worldwide/region.rb
DominiqueFlaaa: nice catch!

DominiqueFlaaa:
feedback addressed! thanks both 🙏

PR #178, opened 2024-05-30
Set specific pnpm version in CI

kennygoff:

Fixes a TS build issue in the CI introduced in #167 due to strict package manager versioning. Also uses correct runner in the npm release job.

See: https://github.com/Shopify/worldwide/actions/runs/9303226439/job/25605062451

Went with making the pnpm version at `9.1.3` in the GH Actions to abide by strict mode.

Alternatively could disable strict mode, if this causes more problems will do that in the future: See https://github.com/pnpm/pnpm/issues/8087

...

...

### Testing

See this PR's passing GH CI build.

### Checklist

### DIFF
--- a/.github/workflows/npm-release.yml
+++ b/.github/workflows/npm-release.yml
@@ -10,7 +10,7 @@ concurrency: ${{ github.workflow }}-${{ github.ref }}
 jobs:
   release:
     name: Release
-    runs-on: shopify-ubuntu-latest
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


gabypancu (approved):

.github/workflows/npm-release.yml
```diff
@@ -10,7 +10,7 @@ concurrency: ${{ github.workflow }}-${{ github.ref }}
 jobs:
   release:
     name: Release
-    runs-on: shopify-ubuntu-latest
+    runs-on: ubuntu-latest
```
devanandersen, line 13: 👍🏼

PR #181, opened 2024-05-30
Additional address concat ts

kennygoff:

Resolves https://github.com/Shopify/address/issues/2576

Implements concatenation functions for the additional address fields for both address1 and address2. These functions will generate a single address string based on one or more sub-fields specified in region yaml config files.

Example:
```ts
// Generate Address1
concatenateAddress1({
  countryCode: 'BR',
  streetName: 'Main',
  streetNumber: '123',
}); // returns 'Main,\u00A0123'
```

Tried to stick closely to the existing Ruby implementation.

Important approach is that I wrote a custom rollup plugin (see `rollup.config.ts`) for the yaml parsing and importing. The [official rollup plugin](https://www.npmjs.com/package/@rollup/plugin-yaml) does not support importing multiple files with one import (and didn't play nicely with other plugins that added that functionality) and one of it's dependencies to generate code was breaking. Luckily it was fairly straightforward, the custom plugin imports all YAMLs using a glob import, uses the same parser that the official plugin used (`js-yaml`), then bundles all the yaml data into one JS object with the country code as the keys for each config. This simplifies usage for us so now we can have a single import that resolves to a map of each config, meaning no TypeScript code updates to add or update regions.

```ts
import regions from 'custom:regions';

region['BR']; // Config for Brazil: { code: 'BR', name: 'Brazil', combined_address_format: {...} }
region['US']; // Config for United States: { code: 'US', name: 'United States' }
```

- `concatenateAddress1.ts` and `concatenateAddress2.ts` and their corresponding tests define how each address's concatenation is defined
- `concatAddressField` in `address-fields.ts` defines the general concatenation logic
- `lang/typescript/README.md` for example usage
- `rollup.config.ts` for package bundling and yaml loading

This PR also sets the npm package to public so changesets will start publishing to NPM after this gets merged.

### Testing

Follow the setup in `lang/typescript/README.md` and run the test suite with `pnpm test`. 

Screenshot 2024-06-03 at 1 25 45 PM

To test the npm package was bundled correctly I also wrote a small test importing the package locally (since its not published yet) to make sure it works and has correct types: 

Screenshot 2024-06-03 at 1 27 33 PM

### Checklist

* [x] Resolve open TODOs
* [x] Fix yaml parsing and subsequent code generation
* [x] Support for all region yaml files (currently manually importing a few examples for tests)


lang/typescript/src/utils.test.ts
```diff
@@ -0,0 +1,113 @@
...
+    };
+    expect(concatAddressField(fieldDefinitionNumberFirst, fieldValues)).toBe(
+      '123\u00A0Main',
+    );
+    expect(concatAddressField(fieldDefinitionNameFirst, fieldValues)).toBe(
+      'Main\u00A0123',
+    );
+  });
+
+  // TODO: Do we care about the required field for concatenation or just when we do actual field validation on the UI/Atlas?
```
kennygoff, line 50: The ruby tests seemed to ignore what was required, seemed like the required field is more for validation, not for the concatenation. Is that right? @DominiqueFlaaa @gabypancu @rochlefebvre

rochlefebvre: Exactly. Systems like Core's checkout negociation and Atlas' own address validation will enforce presence for those fields.  As far as concatenation and splitting goes, all fields are optional.

rochlefebvre:
>fix a parsing issue with US yaml not being compatible with typescript's strict mode

If the problematic parts of the US yaml can be identified and rectified easily, feel free to clean up the file and parse in strict mode.

lang/typescript/src/utils.test.ts
kennygoff: Also cleared up by this change which I'm implementing now: https://github.com/Shopify/worldwide/pull/177

lang/typescript/.changeset/silver-beds-lie.md
josemiguel-alvarez, line 1: Just to get context, why the name of this file?

lang/typescript/README.md
```diff
@@ -1,6 +1,30 @@
 # Worldwide - Typescript
 
-🚧 Still in development -- not ready for use yet 🚧
+Utilities for parsing and formatting address fields
```
josemiguel-alvarez, line 3: The parsing functions are still missing, right?

lang/typescript/.changeset/silver-beds-lie.md
kennygoff: Changesets generates a unique filename with each changelog item so that multiple changesets can be made by multiple people without git clashes. Then they're merged and deleted when a release is created. https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md

lang/typescript/src/extended-address/generateAddress1.ts
```diff
@@ -0,0 +1,35 @@
...
+  address1: Address1,
+): string | null {
+  const config = getRegionConfig(countryCode);
+  const address1CombinedFormat = config?.combined_address_format?.address1;
+  const address1NotEmpty =
+    Object.values(address1).filter((value) => value !== undefined).length > 0;
+
+  if (address1.address1 !== undefined) {
+    // Always default to returning address1 if it's provided
+    return address1.address1;
```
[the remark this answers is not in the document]
kennygoff: @josemiguel-alvarez @DominiqueFlaaa (+ @rochlefebvre who responded to the original thread)

Discussed in standup to make the generate address more strict, such that only `streetName`, `streetNumber` are valid inputs. I think to do this properly we need to exclude any attempts to generate a code outside of ones that have the extended address format defined. This won't change the api for valid inputs (such as Brazil) but will prevent using the function flexibly with other regions.

```ts
// Typescript Error, cannot pass 'US' as a country code
generateAddress1('US', { address1: '123 Main' })
generateAddress1('US', { streetName: 'Main', streetNumber: '123' })
```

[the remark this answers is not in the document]
josemiguel-alvarez: We have the possibility of accepting the full address. Then the library would be responsible of using the correct fields for each country.

lang/typescript/README.md
```diff
@@ -1,6 +1,34 @@
...
+
+## Usage
+
+```ts
+import {generateAddress1, generateAddress2} from '@shopify/worldwide';
+
+// Generate Address1
+generateAddress1({
+  countryCode: 'BR',
+  streetName: 'Main',
```
rochlefebvre, line 13: Nit: can we use a more Brazilian sounding street name?

lang/typescript/README.md
```diff
@@ -1,6 +1,34 @@
...
+  streetNumber: '123',
+}); // returns 'Main\u00A0, 123'
+generateAddress1({
+  countryCode: 'BR',
+  address1: '123 Main',
+}); // returns '123 Main'
+
+// Generate Address2
+generateAddress1({
```
rochlefebvre, line 22: I believe you meant `generateAddress2` here and in the following example.

lang/typescript/src/utils/address-fields.ts
```diff
@@ -0,0 +1,37 @@
...
+  return fieldDefinition.reduce((concatenatedAddress, field, index) => {
+    if (address[field.key]) {
+      // Don't include decorator at the start of the string
+      const decorator =
+        field.decorator && concatenatedAddress.length > 0
+          ? field.decorator
+          : '';
+
+      // Don't include delimeter if only first field is present
+      const delimeter = index === 0 ? '' : RESERVED_DELIMETER;
```
rochlefebvre, line 30: delimiter

lang/typescript/src/utils/address-fields.ts
```diff
@@ -0,0 +1,37 @@
+import {Address} from '../types/address';
+
+import {AdditionalAddressField} from './regions';
+
+export const RESERVED_DELIMETER = '\xA0';
```
rochlefebvre, line 5: RESERVED_DELIMITER

lang/typescript/src/utils/regions.ts
```diff
@@ -0,0 +1,52 @@
+import regions from 'custom:regions';
+
+import {ValidYamlType} from '../types/yaml';
+import {Address} from '../types/address';
+
+export interface AdditionalAddressField {
```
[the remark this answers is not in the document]
kennygoff: Oh yeah forgot to change the type name after we moved to `combined_address_format`.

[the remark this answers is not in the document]
rochlefebvre: Yeah, looking at `CombinedAddressFormat` below, I feel that a combined address format _has one_ **field** and _has many_ **concatenation rules**.

lang/typescript/src/extended-address/generateAddress1.ts
```diff
@@ -0,0 +1,25 @@
+import {concatAddressField} from '../utils/address-fields';
+import {getRegionConfig} from '../utils/regions';
+import {Address} from '../types/address';
+
+/**
+ * Generate a concatenated address1 string based on the region specified by country code
```
rochlefebvre, line 6: Maybe  _based on the region of the given address_ ?  This comment sounds like we're passing in a country code parameter, which was true in an earlier iteration of this code.

lang/typescript/README.md
kennygoff: I took most of the tests/examples from the ruby code which just uses "Main", got a better example?

lang/typescript/src/extended-address/generateAddress1.ts
```diff
@@ -0,0 +1,25 @@
...
+ * @param address Address containing `countryCode` and any number of other
+ * optional address fields
+ * @returns concatenated address string or null if invalid input was sent
+ */
+export function generateAddress1(address: Address): string | null {
+  const config = getRegionConfig(address.countryCode);
+  const address1CombinedFormat = config?.combined_address_format?.address1;
+  const containsAddress1ExtendedFields =
+    address.streetName !== undefined || address.streetNumber !== undefined;
```
rochlefebvre, line 15: Can we generalize this to check that `address` contains at least one of the keys listed in `address1CombinedFormat`? Just so we don't hardcode the assumption that only street names/numbers may appear in address1.

Same recommendation for address2 only containing line2 and neighborhood.

lang/typescript/src/extended-address/generateAddress2.test.ts
```diff
@@ -0,0 +1,150 @@
...
+    expect(
+      generateAddress2({
+        countryCode: 'US',
+        line2: '#2',
+        neighborhood: 'Centretown',
+      }),
+    ).toBeNull();
+  });
+
+  test('returns null if address2 is missing for region with additional fields', () => {
```
rochlefebvre, line 21: returns null if address2, line2 and neighborhood are missing...

lang/typescript/README.md
```diff
@@ -1,6 +1,68 @@
...
+
+```ts
+generateAddress1({
+  countryCode: 'BR',
+  streetName: 'Av. Paulista',
+  streetNumber: '1578',
+}); // returns 'Av. Paulista,\u00A0 1578'
+```
+
+You can generate Address1 or Address2 fields for any supported country, even if
```
rochlefebvre, line 38: I believe generateAddress1 returns null if the country does not define a concatenated format. Maybe we could mention that we return an unmodified address1 value if it is defined in the source address?

rochlefebvre (approved):
The changes look good to me! I am not fluent in javascript, so I recommend that @josemiguel-alvarez also review it.

lang/typescript/src/extended-address/generateAddress1.test.ts
```diff
@@ -0,0 +1,121 @@
...
+      }),
+    ).toBe('Main 123');
+    expect(
+      generateAddress1({
+        countryCode: 'CL',
+        address1: 'Main 123',
+        streetName: 'Main',
+        streetNumber: '123',
+      }),
+    ).toBe('Main 123');
```
josemiguel-alvarez, line 43: I think this is wrong. If a country supports the extended format the additional fields should have higher priority than the standard fields.

With this logic, I have to make sure that every time I use this function I have to remove the previous `address1`. Otherwise the function won't do anything and I'll stay with the old `address1`.

josemiguel-alvarez: > This is expected, yes.

Yes, I understand it was implemented on purpose. But I think it's wrong. 

I think passing a full address so the library takes care of using the correct fields to generate an address1 is ok. But the library is meant to be used with the additional fields and the result that the function generates in this case is wrong. The correct address1 in that case should include the reserved delimiter.

> If there's an existing address1 you're migrating an existing address to the new format it should be split out first.

It doesn't necessarily have to be that we are migrating. Let's say I have the following values (using additional fields):

```js
{
  address1: "Main\u00A0123";
  streetName: "Main";
  streetNumber: "123";
  ...rest
}

```

Now the user writes a new value to the streetNumber field, for example: `456`

I call then this function to generate the new address1 (which I cannot update without calling this function first):

```js
generateAddress1({
  address1: "Main\u00A0123",
  streetName: "Main",
  streetNumber: "456",
  ...rest
});
```

I would get a wrong result. So for me the library wouldn't work as expected. 

If we ship it like this we need to make it sure that developers always clear the `address1` value of the address before calling this function.

josemiguel-alvarez: We need to always store `address1`. Outside of checkout `streetName` and `streetNumber` do not exist. An even inside checkout there are places where they do not exist like wallets or extensions.

kennygoff: Followup after a discussion with @rochlefebvre and @josemiguel-alvarez:

I will modify to follow this set of rules:
- Region does not define `combined_address_field`: return `address1` unmodified (no change)
- Region does define `combined_address_field`, but each extended field is `undefined`: return `address1` unmodified (no change)
- Region does define `combined_address_field`, but any extended field is a string (empty or otherwise): return concatenated address1 (⚠️ change)
- No relevant fields defined: Return `null` (no change)

PR #199, opened 2024-06-05
Generate concatenated address1/2 values when additional field values are present

rochlefebvre:
Updating the ruby counterpart of #181 to align method names and behaviour:

`Worldwide::Address#concatenated_address1` --> `Worldwide::Address#concatenate_address1`
`Worldwide::Address#concatenated_address2` --> `Worldwide::Address#concatenate_address2`

As soon as one of the additional address field values is non-nil, then `concatenate_address1/2` ignores the existing `address1/2` field and returns a concatenated one.

### Testing
See updated AddressTest suite.

...

### Checklist


kennygoff:
Tests look good 👍

PR #140, opened 2024-05-07
Support message format translations

alanhill:
### What are you trying to accomplish?

...

### What approach did you choose and why?

...

### What should reviewers focus on?

...

### The impact of these changes

...

### Testing

...

### Checklist

* [ ] I have added a CHANGELOG entry for this change (or determined that it isn't needed)

PR #205, opened 2024-06-10
All error translations now have a _instructional and _informative message key

bgraham11:

This PR fixes the inconsistent error messaging scheme introduced in https://github.com/Shopify/worldwide/pull/140.

All error messages will now have both have an `_instructional` and `_informative` suffix on their key's. This is asserted in the `FieldDataConsistencyTest`.

If the `Worldwide::Field::error` method is called with a `code` that does not end in either `_instructional` or `_informative`, the `Worldwide::Field::error` method will append `_instructional` to the `code` and attempt to return this message(an empty string is returned if no value is found).

Much simpler as there is no complicated fall back logic. All error translations have both an `_instructional` and `_informative` suffix key for future expansion.

### Checklist
* [x] Bump version following [these steps](https://github.com/Shopify/worldwide/pull/140#issuecomment-2138062266) 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,12 +26,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Security in case of vulnerabilities.
 
 ## [Unreleased]
-- Additional field definitions in AE, CR, KW, PA, PE, SA, updated decorators for PH, VN  [#206](https://github.com/Shopify/worldwide/pull/206)
-- update street_name and street_number too_long error message [#203](https://github.com/Shopify/worldwide/pull/203)
-- Update translations for neighborhood too_long error message, add translations for localized neighborhood errors in MX, PH [#191](https://github.com/Shopify/worldwide/pull/191)
-- Generate concatenated address1/2 values when some additional field values are present [#199](https://github.com/Shopify/worldwide/pull/199)
 
 --
+
+## [1.3.0] - 2024-06-10
+
+- Update translations for neighborhood too_long error message, add translations for localized neighborhood errors in MX, PH [#191](https://github.com/Shopify/worldwide/pull/191)
+- Generate concatenated address1/2 values when some additional field values are present [#199](https://github.com/Shopify/worldwide/pull/199)
+- update street_name and street_number too_long error message [#203](https://github.com/Shopify/worldwide/pull/203)
+- Additional field definitions in AE, CR, KW, PA, PE, SA, updated decorators for PH, VN  [#206](https://github.com/Shopify/worldwide/pull/206)
+- All error messages now contain an `_instructional` and `_informative` key. If the `Worldwide::Field::error` method is called with a `code` that does not end in either `_instructional` or `_informative`, the `Worldwide::Field::error` method will append `_instructional` to the `code` and returns this error message. [#205](https://github.com/Shopify/worldwide/pull/205)
 
 ## [1.2.0] - 2024-06-05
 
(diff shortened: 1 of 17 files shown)


lib/worldwide/field.rb
```diff
@@ -48,6 +48,9 @@ def label_marked_optional(locale: I18n.locale)
     end
 
     def error(locale: I18n.locale, code:, options: {})
+      unless code.end_with?("_instructional") || code.end_with?("_informative")
```
bgraham11, line 51: Reverts logic of this back to https://github.com/Shopify/worldwide/blob/7e4a6892ba83d2183ce811c4563877b2584d46aa/lib/worldwide/field.rb

If a error message key doesn't end with either `_instructional` or `_informative` we will default it to the `_instructional`.

test/worldwide/field_data_consistency_test.rb
```diff
@@ -37,6 +37,21 @@ def sanity_check_file(file_name)
...
           assert_includes(permitted_keys, key)
         end
+
+        assert_instructional_informative_error_messages(value["errors"], file_name, key)
+      end
+    end
+
+    def assert_instructional_informative_error_messages(errors, file_name, field_key)
+      return if errors.nil?
+      return unless file_name.end_with?("en.yml")
```
bgraham11, line 47: This enables `async` translations, while enforcing english as the base language that must exist.

CHANGELOG.md
```diff
@@ -30,7 +30,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Update translations for neighborhood too_long error message, add translations for localized neighborhood errors in MX, PH [#191](https://github.com/Shopify/worldwide/pull/191)
 - Generate concatenated address1/2 values when some additional field values are present [#199](https://github.com/Shopify/worldwide/pull/199)
 
---
+-- All error messages now contain an `_instructional` and `_informative` key. If the `Worldwide::Field::error` method is called with a `code` that does not end in either `_instructional` or `_informative`, the `Worldwide::Field::error` method will append `_instructional` to the `code` and returns this error message.
```
plentz, line 33: I think we need to add a new line with just one `-` and keep the line with the 2 `--`

plentz (approved):
Added one comment, but this is looking great. We should probably release a new version even without the translations (because they we'll likely take a few days to come back)

CHANGELOG.md
bgraham11: Oh shoot! Nice spot.. I have correct this! <3

CHANGELOG.md
```diff
@@ -30,7 +30,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Update translations for neighborhood too_long error message, add translations for localized neighborhood errors in MX, PH [#191](https://github.com/Shopify/worldwide/pull/191)
 - Generate concatenated address1/2 values when some additional field values are present [#199](https://github.com/Shopify/worldwide/pull/199)
 
---
+## [1.3.0] - 2024-10-05
```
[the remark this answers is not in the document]
bgraham11: oooof, woops! I have correct this!

db/data/regions/US/en.yml
```diff
@@ -18,16 +18,17 @@ en:
...
         zip:
           label:
             default: ZIP code
             optional: ZIP code (optional)
           errors:
             blank_instructional: Enter a ZIP code
             blank_informative: ZIP code is missing
-            invalid_for_country_instructional: Enter a valid ZIP code for %{country}
-            invalid_for_country_informative: ZIP code isn't valid for %{country}
+            invalid_for_country_instructional: Enter a valid ZIP code for the %{country}
```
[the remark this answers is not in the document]
bgraham11: This is the override for "United States".

So it reads: `Enter a valid ZIP code for the %{country}`

PR #206, opened 2024-06-10
Add format_extended and additional field defns

DominiqueFlaaa:
Add the format_extended and additional field definitions for all remaining countries in scope 

Also removed any word-decorators, replacing with the default nbsp. 

part of https://github.com/Shopify/address/issues/2616

Added the following keys to the region yml's:
- format_extended
- additional_address_fields
- combined_address_format

Double check the list of countries, and the required and decorator specifications (linked from the original issue) 

When leveraged, will render the extended format for the relevant countries. 

### Testing

...

### Checklist

### DIFF
--- a/db/data/regions/VN.yml
+++ b/db/data/regions/VN.yml
@@ -23,7 +23,7 @@ combined_address_format:
   address2:
     - key: line2
     - key: neighborhood
-      decorator: ", Quận"
+      decorator: ","
 emoji: "\U0001F1FB\U0001F1F3"
 languages:
   - fr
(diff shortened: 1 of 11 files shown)


db/data/regions/VN.yml
```diff
@@ -23,7 +23,7 @@ combined_address_format:
   address2:
     - key: line2
     - key: neighborhood
-      decorator: ", Quận"
```
DominiqueFlaaa, line 26: reflected the key decision to only use punctuation / special characters; no words

gabypancu (approved):

PR #214, opened 2024-06-11
Taiwan support prototype

kennygoff:

Prototype in TS for https://github.com/Shopify/address/issues/2596
...

...

...

...

### Testing

...

### Checklist

### DIFF
--- a/db/data/regions/TW.yml
+++ b/db/data/regions/TW.yml
@@ -27,6 +27,12 @@ combined_address_format:
   address2:
     - key: line2
     - key: neighborhood
+      decorator: ","
+  zh-TW:
+    address2:
+      - key: line2
+      - key: neighborhood
+        delimiter: "\u2060"
 emoji: "\U0001F1F9\U0001F1FC"
 languages:
   - "zh-TW"
--- a/lang/typescript/src/extended-address/concatenateAddress2.test.ts
+++ b/lang/typescript/src/extended-address/concatenateAddress2.test.ts
@@ -182,4 +182,26 @@ describe('concatenateAddress2', () => {
       }),
     ).toBe('\u00A0Centro');
   });
+
+  describe('language override', () => {
+    test("returns default format if language override doesn't match", () => {
+      expect(
+        concatenateAddress2({
+          countryCode: 'TW',
+          line2: 'No. 30',
+          neighborhood: 'Daija District',
+        }),
+      ).toBe('No. 30,\u00A0Daija District');
+    });
+
+    test('returns alternate language format if language override matches', () => {
+      expect(
+        concatenateAddress2({
+          countryCode: 'TW',
+          line2: '30號',
+          neighborhood: '大甲區',
+        }),
+      ).toBe('30號\u2060大甲區');
+    });
+  });
 });
--- a/lang/typescript/src/extended-address/splitAddress2.test.ts
+++ b/lang/typescript/src/extended-address/splitAddress2.test.ts
@@ -37,4 +37,20 @@ describe('splitAddress2', () => {
       neighborhood: 'Centro',
     });
   });
+
+  describe('language override', () => {
+    test("returns default format if language override doesn't match", () => {
+      expect(splitAddress2('TW', 'No. 30,\u00A0Daija District')).toEqual({
+        line2: 'No. 30',
+        neighborhood: 'Daija District',
+      });
+    });
+
+    test('returns alternate language format if language override matches', () => {
+      expect(splitAddress2('TW', '30號\u2060大甲區')).toEqual({
+        line2: '30號',
+        neighborhood: '大甲區',
+      });
+    });
+  });
 });
--- a/lang/typescript/src/utils/address-fields.ts
+++ b/lang/typescript/src/utils/address-fields.ts
@@ -25,7 +25,8 @@ export function concatAddressField(
           : '';
 
       // Don't include delimiter if only first field is present
-      const delimiter = index === 0 ? '' : RESERVED_DELIMITER;
+      const delimiter =
+        index === 0 ? '' : field.delimiter ?? RESERVED_DELIMITER;
 
       return `${concatenatedAddress}${decorator}${delimiter}${address[field.key]}`;
     }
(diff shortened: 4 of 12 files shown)


db/data/regions/TW.yml
```diff
@@ -27,6 +27,12 @@ combined_address_format:
   address2:
     - key: line2
     - key: neighborhood
+      decorator: ","
+  zh-TW:
+    address2:
+      - key: line2
+      - key: neighborhood
+        delimiter: "\u2060"
```
kennygoff, line 35: This nested override looks fine in YAML but when parsing this its really awkward, at least in Typescript.

lang/typescript/src/extended-address/concatenateAddress2.test.ts
```diff
@@ -182,4 +182,26 @@ describe('concatenateAddress2', () => {
...
+    test('returns alternate language format if language override matches', () => {
+      expect(
+        concatenateAddress2({
+          countryCode: 'TW',
+          line2: '30號',
+          neighborhood: '大甲區',
+        }),
+      ).toBe('30號\u2060大甲區');
+    });
+  });
```
kennygoff, line 206: Tests for concat

lang/typescript/src/extended-address/splitAddress2.test.ts
```diff
@@ -37,4 +37,20 @@ describe('splitAddress2', () => {
...
+      });
+    });
+
+    test('returns alternate language format if language override matches', () => {
+      expect(splitAddress2('TW', '30號\u2060大甲區')).toEqual({
+        line2: '30號',
+        neighborhood: '大甲區',
+      });
+    });
+  });
```
kennygoff, line 55: Tests for split

lang/typescript/src/utils/address-fields.ts
```diff
@@ -46,7 +47,13 @@ export function splitAddressField(
...
   concatenatedAddress: string,
 ): Partial<Address> {
-  const values = concatenatedAddress.split(RESERVED_DELIMITER);
+  // Support custom delimiteres defined on ANY sub-field definition
+  const customDelimiter = fieldDefinition.find(
+    (def) => def.delimiter !== undefined,
+  )?.delimiter;
+  const values = concatenatedAddress.split(
+    customDelimiter ?? RESERVED_DELIMITER,
+  );
```
kennygoff, line 56: Here handles the custom delimiter. Based on my current implementation I can't really support multiple delimiters so I'm just getting the first custom one I can find in the language override.

lang/typescript/src/utils/regions.ts
```diff
@@ -25,10 +54,34 @@ function isRegionYamlConfig(
...
+          isYamlObject(value[0]) &&
+          'key' in value[0] &&
+          typeof value[0].key === 'string') ||
+        (key === 'address2' &&
+          isYamlArray(value) &&
+          isYamlObject(value[0]) &&
+          'key' in value[0] &&
+          typeof value[0].key === 'string') ||
+        key === 'zh-TW',
+    )
```
kennygoff, line 84: Can ignore this, this was me _trying_ to strictly type this as a subset of valid yaml and I think this approach might be a fools errand unless I can find use something like [Zod](https://zod.dev/) for it to just write a schema for the region yaml.

The parser will error if it's invalid yaml that's probably Good Enough™️ anyway.

lang/typescript/src/utils/regions.ts
```diff
@@ -46,3 +99,100 @@ export function getRegionConfig(countryCode: string): RegionYamlConfig | null {
...
+      if (value) {
+        return matchesCharset || stringUsesScript(value, script);
+      }
+    }
+
+    return matchesCharset;
+  }, false);
+
+  return charsetFound;
+}
```
kennygoff, line 120: This is the logic to check if any field in the definition matches a particular script ("Han" for Chinese). If we find anything we know we should use the `zh-TW` version.

lang/typescript/src/utils/regions.ts
```diff
@@ -46,3 +99,100 @@ export function getRegionConfig(countryCode: string): RegionYamlConfig | null {
...
+        langScripts[lang],
+      )
+    ) {
+      concatenationRules = overrideConfig[extendedField];
+      break;
+    }
+  }
+
+  return concatenationRules;
+}
```
kennygoff, line 161: This is a lot of nonsense for trying to handle the fact that `combined_address_format` properties can be _either_ the field definition format OR a language that duplicates the layout of `combined_address_format` itself. Its a little messy right now took me a bit of tinkering to get it working.

lang/typescript/src/utils/regions.ts
```diff
@@ -46,3 +99,100 @@ export function getRegionConfig(countryCode: string): RegionYamlConfig | null {
...
+  // Check for a valid language override and use the first one found
+  const langOverrides = Object.keys(langScripts).filter(
+    (lang) => lang in combinedAddressFormat,
+  );
+  for (const lang of langOverrides) {
+    const overrideConfig = config.combined_address_format[lang];
+    if (
+      isLanguageSpecificAddressFormat(overrideConfig) &&
+      isFieldConcatenationRuleArray(overrideConfig[extendedField]) &&
+      stringUsesScript(concatenatedAddress, langScripts[lang])
```
kennygoff, line 190: Same thing as `getConcatenationRules` but just looking at the concatenated string not properties of Address

lang/typescript/src/utils/regions.ts
```diff
@@ -46,3 +99,100 @@ export function getRegionConfig(countryCode: string): RegionYamlConfig | null {
...
+      addressUsesScript(
+        overrideConfig[extendedField],
+        address,
+        langScripts[lang],
+      )
+    ) {
+      concatenationRules = overrideConfig[extendedField];
+      break;
+    }
+  }
```
kennygoff, line 158: This is the important bit that basically does this:

1. Get all languages overrides defined on this config (in our Taiwan case we should find one, `zh-TW`)
2. For each override, call `addressUsesScript` and check if the Address has a matching character set for the lanuage specified in the override
   - Matching characters found? Use that override config instead (`zh-TW` config for our example)
   - No matching characters? Use the normal `combined_address_format`

PR #215, opened 2024-06-12
add localized error strings for neighborhood

DominiqueFlaaa:
Add localized neighborhood error strings for remaining countries in scope (linked in issue below). 

part of https://github.com/Shopify/address/issues/2616

Copied existing error codes from TR.yml, pasted and updated the regionalized name. 

Double check that all countries in the sheet (in issue) are covered. 

The neighborhood error messages will be regionalized, replacing `neighborhood` with District, Block, or Area when relevant. 

### Testing

...

### Checklist


db/data/regions/TW/en.yml
```diff
@@ -10,7 +10,7 @@ en:
           errors:
             blank_instructional: Enter a district
             blank_informative: District is missing
-            too_long_instructional: District is too long (maximum is 255 characters)
+            too_long_instructional: District is too long (maximum is %{limit} characters)
```
DominiqueFlaaa, line 13: Updating to use the limit param; matches the default message.

gabypancu (approved):

PR #216, opened 2024-06-12
Multi-language support for Taiwan

kennygoff:

Related to https://github.com/Shopify/address/issues/2596

**Updated** After the pairing on the Ruby implementation, @gabypancu and I changed the formatting of the yaml to avoid some of the issues I was having in TS with parsing. Mainly we put everything under `combined_address_format.default`, and changed from setting language as `zh-TW` to setting script as `Latin`, so we didn't have to include a map of languages to unicode scripts.

```yaml
combined_address_format:
  default:
    address2:
      - key: line2
      - key: neighborhood
  Latin:
    address2:
      - key: line2
      - key: neighborhood
        decorator: ","
```

<details><summary>Original version</summary>

Updates the Taiwan region yaml to include alternate `combined_address_format` for Chinese with `zh-TW` language override. When Han script is detected in the address we use the alternate format, which just removes the decorator.

```yaml
combined_address_format:
  address2:
    - key: line2
    - key: neighborhood
      decorator: ","
  zh-TW:
    address2:
      - key: line2
      - key: neighborhood
```

</details> 

TODO:
- [x] Ruby implemenation
- [ ] Decide on delimiter (maybe a separate PR)
- [ ] Decide on field ordering (maybe a separate PR)

**Updated** Started to move away from this to simplify the code to address some of the complexities found with my TypeScript implementation.

~Opted for the language-specific definition in the yaml to be nested and when found for a given field and will completely overwrite the default. This adds a small amount of duplication since `address2` is listed twice, but it's explicit so it's simple to read and means we can swap out whole parts of the format, like we are doing with removing the decorator.~

~This is a little awkward in Typescript, since the `combined_address_format` is now nested, we don't have as clean of a delineation with the types expected for `combined_address_format`'s properties, requiring some extra typeguarding. I may rework this to separate the language specific config into it's own object in TS, while leaving it as-is in the yaml.~

- The config changes in `TW.yml` and addition of `default` property
- Script detection
  - TS: in `stringExclusively UsesScript` and `addressUsesScript` in `regions.ts`
  - Ruby: beginning of split/concat functions `detected_scripts` in `address.rb`

Also left some comments on specific parts of the code

...

### Testing

For typescript run `pnpm test` within `lang/typescript`:

Screenshot 2024-06-12 at 11 30 42 AM

### Checklist

### DIFF
--- a/lib/worldwide/address.rb
+++ b/lib/worldwide/address.rb
@@ -534,9 +534,5 @@ module Worldwide
       delimiter = next_field_definition&.fetch("decorator", nil) || ""
       current_field_value&.delete_suffix(delimiter)
     end
-
-    def detect_script(str)
-      str
-    end
   end
 end


lang/typescript/src/utils/script.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const scriptRegexes: Record<Script, RegExp> = {
+  Arabic: /\p{Script=Arabic}/u,
+  Han: /\p{Script=Han}/u,
+  Hangul: /\p{Script=Hangul}/u,
+  Hiragana: /\p{Script=Hiragana}/u,
+  Katakana: /\p{Script=Katakana}/u,
+  Latin: /\p{Script=Latin}/u,
+  Thai: /\p{Script=Thai}/u,
+};
```
kennygoff, line 20: I took this list from https://github.com/Shopify/worldwide/blob/main/lib/worldwide/scripts.rb, but there's actually a bunch more Regex supports: https://www.regular-expressions.info/unicode.html#script. Not sure if the ones in ruby are the subset we support or something else...

lib/worldwide/address.rb
```diff
@@ -147,7 +147,14 @@ def valid?
 
     RESERVED_DELIMITER = "\u00A0"
     def concatenate_address1
-      additional_fields = region.combined_address_format["address1"] || []
+      detected_scripts = Worldwide::Scripts.identify(text: street_name).map(&:to_s)
```
kennygoff, line 150: Right now in ruby we're only doing this check on `street_name`, but in typescript we're doing it on any field defined in `address1`.

lib/worldwide/address.rb
```diff
@@ -156,7 +163,14 @@ def concatenate_address1
     end
 
     def concatenate_address2
-      additional_fields = region.combined_address_format["address2"] || []
+      detected_scripts = Worldwide::Scripts.identify(text: neighborhood).map(&:to_s)
```
kennygoff, line 166: Right now in ruby we're only doing this check on `neighborhood`, but in typescript we're doing it on any field defined in `address2`.

test/worldwide/address_test.rb
```diff
@@ -172,6 +172,20 @@ class AddressTest < ActiveSupport::TestCase
...
+      assert_equal "No. 3, Daija District", address.concatenate_address2
+    end
+
+    test "concatenate_address2 returns a concatenated string with script-specific formatting - default" do
+      address = Address.new(line2: "30號", neighborhood: "大甲區", country_code: "TW")
+
+      assert_equal "30號 大甲區", address.concatenate_address2
+    end
+
+    # add test for mixed scripts
```
kennygoff, line 187: TODO!

lib/worldwide/address.rb
rochlefebvre: Seeing as both line2 and neighborhood are optional, I think we should consider running the identification on both fields, perhaps prefering the latter? Unit numbers may not be the best predictors, but they could act as tiebreakers.

PR #221, opened 2024-06-14
Change delimiter from Non-Breaking Space to Word Joiner

kennygoff:

Resolves https://github.com/Shopify/address/issues/2630

Change the reserved delimiter from the Non-Breaking Space (`U+00A0`) to Word Joiner (`U+2060`) and updated region decorators to include a space explicitly where necessary.

The Word Joiner character has a few benefits:
- Has zero-width so doesn't add spacing in languages that don't normally have spacing between characters (e.g. Mandarin Chinese)
- Does not introduce a leading spacing when optional address fields are blank
   - Example: Brazil address2 that includes neighborhood without line2 would render as `" Neighborhood"`, will now render as `"Neighborhood"`)

The Ruby and Typescript implementations should handle this change the same way:
- Reserved delimiter updated to `U+2060`
- 2-character decorators need to work, so `", "` for Brazil
- Decorators should be inserted before the next fields delimiter, so `<line2><decorator><delimiter><neighborhood>` for Brazil will be `dpto 4, \u2060Centro`

No changes to the exposed functions or classes.

Internally:
- Decorators are now required on most regions to insert a space

Behavioral/UX change (out of scope of this package but a downstream impact on usage):
- Copy-paste behavior in browsers will change when rendering the concatenated string directly to HTML because Chrome will convert NBSP to a space but will not remove WordJoiner

### Typescript ✅ 

Screenshot 2024-06-14 at 1 45 32 PM

### Ruby 💎 
- tests updated
- some tests added to cover decorator-less countries (TW)
- added a consistency check to ensure that all `decorator` entries in region ymls are either null or non-empty strings.

### Checklist

### DIFF
--- a/test/worldwide/address_test.rb
+++ b/test/worldwide/address_test.rb
@@ -94,20 +94,20 @@ module Worldwide
       cl_address = Address.new(street_number: "123", country_code: "CL")
       br_address = Address.new(street_number: "123", country_code: "BR")
 
-      assert_equal " 123", cl_address.concatenate_address1
-      assert_equal " 123", br_address.concatenate_address1
+      assert_equal "⁠123", cl_address.concatenate_address1
+      assert_equal "⁠123", br_address.concatenate_address1
     end
 
     test "concatenate_address1 returns street name concatenated with street number separated by delimiter" do
       address = Address.new(street_name: "Main Street", street_number: "123", country_code: "CL")
 
-      assert_equal "Main Street 123", address.concatenate_address1
+      assert_equal "Main Street ⁠123", address.concatenate_address1
     end
 
     test "concatenate_address1 returns street name concatenated with street number separated by delimiter and decorator" do
       address = Address.new(street_name: "Main Street", street_number: "123", country_code: "BR")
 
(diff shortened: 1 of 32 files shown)


test/worldwide/address_test.rb
```diff
@@ -94,20 +94,23 @@ class AddressTest < ActiveSupport::TestCase
...
 
-      assert_equal " 123", cl_address.concatenate_address1
-      assert_equal " 123", br_address.concatenate_address1
+      assert_equal "⁠123", cl_address.concatenate_address1
+      assert_equal "⁠123", br_address.concatenate_address1
     end
 
     test "concatenate_address1 returns street name concatenated with street number separated by delimiter" do
-      address = Address.new(street_name: "Main Street", street_number: "123", country_code: "CL")
+      Worldwide::Region.any_instance.stubs(:combined_address_format).returns(
```
rochlefebvre, line 102: All regions with address1 concatenation rules define a decorator. I'm using stubbing to define one without a decorator.

gabypancu (approved):
Nice to see those leading spaces disappear 👏

PR #253, opened 2024-07-09
Respect `available_locales` if already set in application.

Schwad:
Currently, this gem overrides if `config.i18n.available_locales` is manually set in an initializer or environment file. This PR respects if that config is already set.

### Checklist


najibna (commented):
Skipping `available_locales=` when the app already set it also skips `expanded_locales_from_configuration`. Apps that set `[:en, :fr]` would then lose the descendant locales Worldwide currently adds for fallbacks.

PR #258, opened 2024-07-23
Add field consistency test for optional labels 

DominiqueFlaaa:
Copied 
Noticed in a recent translations PR, https://github.com/Shopify/worldwide/pull/256/files, that the suggested change only applies to the `default` label and not the corresponding `optional` label. 

Screenshot 2024-07-23 at 12 08 58 PM

This will result in a slow drift of the optional / default field labels overtime, where the contents of the strings do not match. 

Copied the test from country_db - https://github.com/Shopify/country_db/blob/1574e852a86b8215b221152c24c803f1fee7c36e/test/lib/db_consistency_test.rb#L878

Went through each failing test and updated the label to the correct version. 

...

Ensures the labels stay in sync. 
...

### Testing

...

### Checklist


db/data/regions/CR/el.yml
```diff
@@ -6,7 +6,7 @@ el:
         province:
```
DominiqueFlaaa, line 6: Costa Rica has provinces

db/data/regions/TH/el.yml
```diff
@@ -5,7 +5,7 @@ el:
       addresses:
         province:
           label:
-            default: Νομός
+            default: Επαρχία
             optional: Επαρχία (προαιρετικά)
```
DominiqueFlaaa, line 9: Thailand has province, province in Greek is `Επαρχία`

db/data/regions/TH/hr.yml
```diff
@@ -6,7 +6,7 @@ hr:
         province:
```
DominiqueFlaaa, line 6: Thailand has provinces. Province in Croatian is `Pokrajina`

plentz (approved):

PR #259, opened 2024-07-24
Update iana_to_rails mappings for rails-supported timezones

DominiqueFlaaa:
### Context
The Admin loads the list of timezones from via Atlas. 
Atlas calls [Shopifyi18n which uses](https://github.com/Shopify/shopify-i18n/blob/3af70facf586a9e4b1ad6f01aa959d4a85b913cc/lib/shopify-i18n/time_zone.rb#L12-L16) ActiveSupport::TimeZone#all ([ref](https://api.rubyonrails.org/v7.1.3.4/classes/ActiveSupport/TimeZone.html#method-c-all)) to generate the list of timezones.
[ActiveSupport::TZ loads the timezones](https://github.com/rails/rails/blob/19eebf6d33dd15a0172e3ed2481bec57a89a2404/activesupport/lib/active_support/values/time_zone.rb#L208C8-L208C25) from the [TZInfo gem](https://github.com/tzinfo/tzinfo), which uses files generated from the [IANA Time Zone Database](https://www.iana.org/time-zones).

The list of [timezone mappings](https://github.com/rails/rails/blob/19eebf6d33dd15a0172e3ed2481bec57a89a2404/activesupport/lib/active_support/values/time_zone.rb#L90) shows that Rails support the Europe/Amsterdam timezone, ex see t

Worldwide has a list of iana_to_rails_time_zone mappings which maps the IANA timezones to the rails equivalent. This file maps the iana names to the supported rails timezone names. 

**Hypothesis: Some of the iana_to_rails_time_zone keys are _already supported_ by rails, but are mapped to alternative / linked timezones.** 
This would explain why the option is shown in the admin (via the atlas time_zone endpoint), and when selected by the user, is instead mapped to the linked equivalent. 

### Investigation

Comparing this [iana_to_rails yml](https://github.com/Shopify/worldwide/blob/7b617082c60f689c719f4b8d89a361e47498f8d3/db/data/iana_to_rails_time_zone.yml) and the [timezone/en.yml](https://github.com/Shopify/worldwide/blob/1755907fa136c6d133abdf65b3b69db4b4eff6b4/data/other/timezones/en.yml) (rails supported timezones) shows the following keys are present across both files:
```
America/Guatemala
America/Guyana
America/Monterrey
Asia/Kuwait
Asia/Muscat
Europe/Amsterdam
Europe/Ljubljana
Europe/Sarajevo
Europe/Skopje
Europe/Zagreb
Pacific/Guadalcanal
```

related to https://github.com/Shopify/address/issues/2673

Updated the iana_to_rails mapping, updating the keys that are already supported in rails. The keys now map back to themselves; following the precedent set by other rails-supported key-value pairs: 

https://github.com/Shopify/worldwide/blob/6e77c9db33264b4a0430b0e83a7ffc277247b106/db/data/iana_to_rails_time_zone.yml#L175

...

...

### Testing

...

### Checklist


db/data/iana_to_rails_time_zone.yml
```diff
@@ -101,10 +101,10 @@ America/Ciudad_Juarez: America/Denver
 America/Coral_Harbour: America/Bogota
 America/Cordoba: America/Argentina/Buenos_Aires
 America/Costa_Rica: America/Mexico_City
-America/Creston: America/Phoenix  
+America/Creston: America/Phoenix
 America/Cuiaba: America/Caracas
 America/Curacao: America/Puerto_Rico
-America/Danmarkshavn: Etc/Utc
+America/Danmarkshavn: Etc/UTC
```
DominiqueFlaaa, line 107: This incorrect value was flagged by one of the new tests. The supported timezone is `Etc/UTC`, [ref](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones)
```

Failure:
Worldwide::TimeZoneDataConsistencyTest#test_iana_to_rails_time_zone_keys_map_to_an_equivalent_value_when_supported_by_rails [/Users/dominiquefl/src/github.com/Shopify/worldwide/test/worldwide/time_zone_data_consistency_test.rb:37]
Minitest::Assertion: Etc/Utc timezone is not supported by Rails.
```

test/worldwide/deprecated_time_zone_mapper_test.rb
```diff
@@ -48,602 +48,13 @@ class TimeZoneTest < ActiveSupport::TestCase
     def iana_data_input_expected
       [
         ["Africa/Abidjan", "Africa/Monrovia"],
-        ["Africa/Accra", "Africa/Monrovia"],
```
DominiqueFlaaa, line 51: This test is brittle; updated to check a few of the expected cases.

skelley-ton:
> Code changes look fine to me but we should await @pegelston's review before merging as he's got the context around the initial work here.

I did the original changes, when I created the 600 mappings by hand I must have messed some up so I think we are good to ship @DominiqueFlaaa

PR #261, opened 2024-07-25
Add Kosovo to world.yml

benkovy:

Atlas leverages worldwide but is unable to resolve the `associated_continent` for Kosovo. Because of this, we fallback to Europe which is not translated.

### Checklist


DominiqueFlaaa (approved):

DominiqueFlaaa:
could you update the changelog with this change?

PR #281, opened 2024-08-30
Add option to fall back on regex parsing to splitAddress1

gabypancu:

Resolves https://github.com/Shopify/address/issues/2726

When there is no reserved delimiter present in an address line (i.e. any address that was created before the release of additional address fields), we want to attempt to split the additional fields out using regex. The regex patterns will have to be strict in order to minimize incorrect captures.

1. Adds a conservative address1 regex pattern for the Netherlands. On a sample of about 800 address lines, it parses 75% correctly (the rest are not matched) - https://rubular.com/r/3BNDUviIL8hHeS
~~2. Adds a `splitAddress1WithFallback` function to the TS library that uses the regex as a fallback if splitting on the delimiter is not possible~~
2. Adds an optional param `tryRegexFallback` to the existing `splitAddress1` function in the TS library. When true, and when the input string does not contain the reserved delimiter, we attempt to parse using regex.

This way, checkout web still only has one function to call for splitting.

The following PRs have been merged into this one:
- https://github.com/Shopify/worldwide/pull/282
- https://github.com/Shopify/worldwide/pull/286

**See PR comments for debate on the following pros and cons**
Our options for where to implement this new behaviour:
1. In the existing `splitAddress1` function, controlled by a new optional param
- Requires the smallest change in checkout-web, but brings higher visibility to this feature
- Reusability if we choose to use this on surfaces other than checkout-web
2. In a new function in Worldwide
- Maintains separation with the existing splitting function, reduces visibility to this feature
- Reusability if we choose to use this on surfaces other than checkout-web
3. Directly in checkout-web when splitting fails
- Keeps this feature private to Shopify
- Introduces complexity & Worldwide domain into checkout-web
- Not reusable outside of checkout-web

### Checklist

### DIFF
--- a/db/data/regions/NL.yml
+++ b/db/data/regions/NL.yml
@@ -30,6 +30,8 @@ format:
   show: "{firstName} {lastName}_{company}_{address1}_{address2}_{zip} {city}_{country}_{phone}"
 format_extended:
   edit: "{country}_{firstName}{lastName}_{company}_{streetName}{streetNumber}_{address2}_{zip}{city}_{phone}"
+address1_regex:
+  - "^(?<streetName>[^\\d]+) (?<streetNumber>\\d+(?: ?[a-z])?)$"
 additional_address_fields:
   - name: streetName
     required: true
--- a/lang/typescript/CHANGELOG.md
+++ b/lang/typescript/CHANGELOG.md
@@ -1,5 +1,31 @@
 # @shopify/worldwide
 
+## 0.7.0
+
+### Minor Changes
+
+- bbacde1: Add address1 regex for BE, BR, CL, ES, IL, MX
+- b6e1c7f: Add address1 regex for DE
+- ee7b7a9: Add optional tryRegexFallback param to splitAddress1 function to attempt splitting address lines that do not contain the reserved delimiter
+
+## 0.7.0-next.2
+
+### Minor Changes
+
+- 3bfb56a: Add address1 regex for DE
+
+## 0.7.0-next.1
+
+### Minor Changes
+
+- Add address1 regex for BE, BR, CL, ES, IL, MX
+
+## 0.7.0-next.0
+
+### Minor Changes
+
+- 1f5d405: Add optional tryRegexFallback param to splitAddress1 function to attempt splitting address lines that do not contain the reserved delimiter
+
 ## 0.6.0
 
 ### Minor Changes
--- a/lang/typescript/src/extended-address/splitAddress1.test.ts
+++ b/lang/typescript/src/extended-address/splitAddress1.test.ts
@@ -2,35 +2,382 @@ import {splitAddress1} from './splitAddress1';
 
 describe('splitAddress1', () => {
   test('returns null when extended address is not defined for region', () => {
-    expect(splitAddress1('US', '123 Main')).toBeNull();
+    expect(splitAddress1('US', '123 Main', false)).toBeNull();
+    expect(splitAddress1('US', '', false)).toBeNull();
+    expect(splitAddress1('US', '123 Main', true)).toBeNull();
+    expect(splitAddress1('US', '', true)).toBeNull();
   });
 
   test('returns empty object when extended address string is empty', () => {
-    expect(splitAddress1('CL', '')).toEqual({});
-    expect(splitAddress1('BR', '')).toEqual({});
+    expect(splitAddress1('CL', '', false)).toEqual({});
+    expect(splitAddress1('BR', '', false)).toEqual({});
+    expect(splitAddress1('CL', '', true)).toEqual({});
+    expect(splitAddress1('BR', '', true)).toEqual({});
   });
 
-  test('returns address1 as street name when no delimiter is present', () => {
-    expect(splitAddress1('CL', '123 Main')).toEqual({streetName: '123 Main'});
-    expect(splitAddress1('BR', '123 Main')).toEqual({streetName: '123 Main'});
+  test('returns address1 as street name when no delimiter is present and tryRegexFallback is false', () => {
+    expect(splitAddress1('CL', '123 Main', false)).toEqual({
+      streetName: '123 Main',
+    });
+    expect(splitAddress1('BR', '123 Main', false)).toEqual({
+      streetName: '123 Main',
+    });
   });
+
+  test('returns address1 as street name when no delimiter is present, tryRegexFallback is true, and regex is not defined', () => {
+    expect(splitAddress1('CL', '123 Main', true)).toEqual({
+      streetName: '123 Main',
+    });
+    expect(splitAddress1('BR', '123 Main', true)).toEqual({
+      streetName: '123 Main',
+    });
+  });
+
+  test.each([
+    {
+      country: 'NL',
+      address: 'Kempenaar 25 11',
+      expected: {streetName: 'Kempenaar 25 11'},
+    },
+    {
+      country: 'NL',
+      address: '40 Baandersstraat',
+      expected: {streetName: '40 Baandersstraat'},
+    },
+    {
+      country: 'BR',
+      address: 'Main, 123, Apt 2',
+      expected: {streetName: 'Main, 123, Apt 2'},
+    },
+  ])(
+    'returns address1 as street name when no delimiter is present, tryRegexFallback is true, and address does not match regex',
+    ({country, address, expected}) => {
+      expect(splitAddress1(country, address, true)).toEqual(expected);
+    },
+  );
 
   test('returns street number if string before delimiter is empty', () => {
-    expect(splitAddress1('CL', '\u2060123')).toEqual({streetNumber: '123'});
-    expect(splitAddress1('BR', '\u2060123')).toEqual({streetNumber: '123'});
+    expect(splitAddress1('CL', '\u2060123', false)).toEqual({
+      streetNumber: '123',
+    });
+    expect(splitAddress1('BR', '\u2060123', false)).toEqual({
+      streetNumber: '123',
+    });
+    expect(splitAddress1('CL', '\u2060123', true)).toEqual({
+      streetNumber: '123',
+    });
+    expect(splitAddress1('BR', '\u2060123', true)).toEqual({
+      streetNumber: '123',
+    });
   });
 
   test('returns full address object when separated by delimiter', () => {
-    expect(splitAddress1('CL', 'Main \u2060123')).toEqual({
+    expect(splitAddress1('CL', 'Main \u2060123', false)).toEqual({
+      streetName: 'Main',
+      streetNumber: '123',
+    });
+    expect(splitAddress1('CL', 'Main \u2060123', true)).toEqual({
       streetName: 'Main',
       streetNumber: '123',
     });
   });
 
   test('returns full address object when separated by delimiter and decorator', () => {
-    expect(splitAddress1('BR', 'Main, \u2060123')).toEqual({
+    expect(splitAddress1('BR', 'Main, \u2060123', false)).toEqual({
+      streetName: 'Main',
+      streetNumber: '123',
+    });
+    expect(splitAddress1('BR', 'Main, \u2060123', true)).toEqual({
       streetName: 'Main',
       streetNumber: '123',
     });
   });
+
+  test.each([
+    {
+      address: 'Mercuriusstraat 26',
+      expected: {streetName: 'Mercuriusstraat', streetNumber: '26'},
+    },
+    {
+      address: 'Bloemgracht 41B',
+      expected: {streetName: 'Bloemgracht', streetNumber: '41B'},
+    },
+    {
+      address: 'Bloemgracht 41b',
+      expected: {streetName: 'Bloemgracht', streetNumber: '41b'},
+    },
+    {
+      address: 'Meester Arendstraat 48 B',
+      expected: {streetName: 'Meester Arendstraat', streetNumber: '48 B'},
+    },
+  ])(
+    'returns full address object when not separated by delimiter, tryRegexFallback is true and address matches regex for NL',
+    ({address, expected}) => {
+      expect(splitAddress1('NL', address, true)).toEqual(expected);
+    },
+  );
+
+  test.each([
+    {
+      address: 'Ziegeleiweg 3',
+      expected: {streetName: 'Ziegeleiweg', streetNumber: '3'},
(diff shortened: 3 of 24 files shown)


db/data/regions/NL.yml
```diff
@@ -30,6 +30,7 @@ format:
   show: "{firstName} {lastName}_{company}_{address1}_{address2}_{zip} {city}_{country}_{phone}"
 format_extended:
   edit: "{country}_{firstName}{lastName}_{company}_{streetName}{streetNumber}_{address2}_{zip}{city}_{phone}"
+address1_regex: "^(?<streetName>[^\\d]+)\\s(?<streetNumber>\\d+(?:\\s?[A-Za-z])?)$"
```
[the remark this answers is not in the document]
rochlefebvre: `\s` always matches the space, tab, newline, carriage return, and form feed characters. I believe Javascript also includes unicode flavours as well.

How about we be more restrictive and just match on literal space characters?

```regex
^(?<streetName>[^\\d]+) (?<streetNumber>\\d+(?: ?[A-Za-z])?)$
```

lang/typescript/src/extended-address/splitAddress1WithFallback.test.ts
```diff
@@ -0,0 +1,80 @@
...
+      },
+    ];
+
+    testCases.forEach(({input, expected}) => {
+      expect(splitAddress1WithFallback(input.country, input.address)).toEqual(
+        expected,
+      );
+    });
+  });
+});
```
kennygoff, line 80: For this and the above test that handles multiple test cases Vitest has a utility called [`test.each`](https://vitest.dev/api/#test-each). You can pass the inputs as an object to `test.each` and provide a clean way to print the results for each test case. I'm not sure if you can pass objects as input though so you might have to do flatten the input objects:

```ts
// Nested objects might work, but I'm not 100% sure
test.each([
  { input: { country: 'NL', address: 'Kempenaar 25 11' }, expected: 'Kempenaar 25 11' }
])('splitAddress1WithFallback($input.country, $input.country) -> $expected', ({ input, expected }) => {
  expect(splitAddress1WithFallback(input.country, input.address)).toBe(expected)
})

// Flattened, definitely will work
test.each([
  { country: 'NL', address: 'Kempenaar 25 11', expected: 'Kempenaar 25 11' }
])('splitAddress1WithFallback($country, $address) -> $expected', ({ country, address, expected }) => {
  expect(splitAddress1WithFallback(country, address)).toBe(expected)
})
```

lang/typescript/src/utils/address-fields.ts
```diff
@@ -34,8 +34,9 @@ export function concatAddressField(
 }
 
 /**
- * Utility function that parses an address string based on a provided field
- * definition and set of values for those fields
+ * Utility function that splits an address string based on a provided field
+ * definition and set of values for those fields, with a fallback that uses a
+ * regex pattern
```
kennygoff, line 39: This function wasn't updated so this doc description shouldn't change

lang/typescript/src/extended-address/splitAddress1WithFallback.ts
```diff
@@ -0,0 +1,35 @@
...
+ * country code, fall back on an address1 regex pattern
+ * @param countryCode 2-letter country code string
+ * @param concatenatedAddress Combined address1 string
+ * @returns Partial Address object containing split address fields or null if
+ * the region does not define an extended address format
+ */
+export function splitAddress1WithFallback(
+  countryCode: string,
+  concatenatedAddress: string,
+): Partial<Address> | null {
```
kennygoff, line 20: @gabypancu I actually think I prefer just changing the default behavior or making it an optional param to turn on/off the regex fallback. Not a super strong opinion so if you have a good reason this is fine, but the library is new enough and only has one main consumer so I feel like making the decision now to change it to add this fallback always if the user wants to opt-out. At the end of the day all of the split logic is defined in the region yamls so I think we can safely change this and just bump the major version to signify the clear change in scope.

I don't hate just doing `splitAddress1('NL', 'Main street 123', false)` even if its a little clunky if we want to offer a way to turn off the fallback regex.

If we go with a second function we should document this in the README too.

db/data/regions/NL.yml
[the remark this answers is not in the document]
gabypancu: Good callout 👍 done!

lang/typescript/src/utils/address-fields.ts
```diff
@@ -71,3 +72,33 @@ export function splitAddressField(
 
   return parsedAddressObject;
 }
+
+/**
+ * Utility function that parses an address string based on a provided field
+ * definition and set of values for those fields, and a regex pattern as a fallback
+ *
+ * @param regex Regex pattern for parsing out sub-fields
```
kennygoff, line 80: Missing the first param for fieldDefinition here `@param fieldDefinition Array of definitions of address sub-fields`

kennygoff (changes requested):

lang/typescript/src/extended-address/splitAddress1WithFallback.ts
gabypancu: This was initially my preference as well. @rochlefebvre brought up a good point that this "fallback" logic is not ideal, and we don't want to draw attention to it. That was the main motivator for having it in a separate function.

lang/typescript/src/utils/address-fields.ts
gabypancu: Thanks for catching that!

lang/typescript/src/extended-address/splitAddress1WithFallback.ts
kennygoff: Yeah just run the pnpm changeset is all. I was recommending to update the README to document usage of the new function but if we want it to be undocumented on purpose then no need. After it gets merged to main a PR will open to handle the release.

kennygoff: I think my main concern right now is that this is still a public facing function but it's not really clear why it's there when reading through it without context. So if someone goes to import the documented `splitAddress1` function they will see a semi-documented second `splitAddress1WithFallback` function with a similar name in autocomplete/IntelliSense. I think this might have the opposite effect we want: "Oh I should use the one with the fallback, because it has fallback functionality which is a good thing". We could double-down on the "don't use this" or "it's experimental" and just popping a warning in the docstring comment for the new fallback function.

Imho though I think adding it as an explicit set of options on the existing function by doing `splitAddress1('NL', 'Main 123', { experimentalRegexFallback: true })`. That way it's opt-in and very clear it's potentially not something someone should use without knowing what they're doing. Then if don't want to document it in the README or later we decide we want to no longer expose this option on the public surface of the API we can just remove it.

gabypancu: Discussed off-Github - we've aligned on moving the fallback logic into the existing `splitAddress1` function behind an optional `experimentalRegexFallback` flag for the reasons outlined above. We will also release this as a beta feature, given that support for countries will be added incrementally

lang/typescript/src/extended-address/splitAddress1.ts
```diff
@@ -1,25 +1,52 @@
...
+ * @param address Combined address1 string
+ * @param experimentalRegexFallback Flag to attempt regex parsing as a fallback mechanism
  * @returns Partial Address object containing parsed address fields or null if
  * the region does not define an extended address format
  */
 export function splitAddress1(
   countryCode: string,
-  concatenatedAddress: string,
+  address: string,
+  experimentalRegexFallback = false,
```
gabypancu, line 28: Open to other names. I considered `tryRegexFallback` as it's more concise and suggests that the splitting may not be successful

lang/typescript/src/extended-address/splitAddress1WithFallback.test.ts
gabypancu: Oh nice! Applied 👍

lang/typescript/src/extended-address/splitAddress1.ts
kennygoff: `try` as a prefix works, although I think `experimental` feels more "dangerous" which seems like it's what we want to convey. Up to you, I'm not too concerned with the specific prefix though as long as it seems to make sense when reading. Just don't use `useRegexFallback` as that prefix has a specific use case in react and could confuse consumers.

lang/typescript/src/extended-address/splitAddress1.ts
```diff
@@ -1,25 +1,52 @@
...
 ): Partial<Address> | null {
   const config = getRegionConfig(countryCode);
   const fieldConcatenationRules = config
-    ? getConcatenationRules(config, concatenatedAddress, 'address1')
+    ? getConcatenationRules(config, address, 'address1')
     : undefined;
+  const address1Regex = config ? getAddress1Regex(config) : undefined;
+  if (address === '') {
+    return {};
+  }
```
kennygoff, line 37: This will change existing behavior for parsing on countries that do not have a config thats untested right now. I'll suggest a code edit to include a test case to catch this behavior.

```ts
splitAddress1('US', '')) // returns null today
splitAddress1('US', '')) // returns {} with this change
```

We might want to invert this big if statement below and just return early:

```ts
if(!fieldConcatenationRules) {
  return null;
}

if (address === '') {
  return {};
}

// Now do the rest of the important config checks not in an if block
```

lang/typescript/src/extended-address/splitAddress1.test.ts
```diff
@@ -3,34 +3,111 @@ import {splitAddress1} from './splitAddress1';
 describe('splitAddress1', () => {
   test('returns null when extended address is not defined for region', () => {
     expect(splitAddress1('US', '123 Main')).toBeNull();
+    expect(splitAddress1('US', '123 Main', true)).toBeNull();
```
kennygoff, line 6: The current behavior is to return null if the country has no region config, regardless of what is inputted. Here's some new test cases that'll cover that.

```suggestion
    expect(splitAddress1('US', '123 Main')).toBeNull();
    expect(splitAddress1('US', '')).toBeNull();
    expect(splitAddress1('US', '123 Main', true)).toBeNull();
    expect(splitAddress1('US', '', true)).toBeNull();
```

lang/typescript/src/utils/address-fields.ts
```diff
@@ -71,3 +71,26 @@ export function splitAddressField(
...
+
+/**
+ * Utility function that attempts to parse an address string based on a regex
+ *
+ * @param fieldDefinition Array of definitions of address sub-fields
+ * @param regexPatterns Regex patterns for parsing sub-fields from an address string
+ * @param address Address string to parse
+ * @returns Partial Address object of fields parsed from string
+ */
+export function regexSplitAddressField(
```
kennygoff, line 83: Can you also add a test for this function in `address-fields.test.ts`?

lang/typescript/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # @shopify/worldwide
 
+## 0.7.0-next.0
+
+### Minor Changes
+
+- 1f5d405: Add optional experimentalRegexFallback param to splitAddress1 function to attempt splitting address lines that do not contain the reserved delimiter
+
```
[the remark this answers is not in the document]
kennygoff: Looks like it was but I don't see the package published as a prerelease on NPM so might have to change something in the GH Actions to actually do it https://www.npmjs.com/package/@shopify/worldwide?activeTab=versions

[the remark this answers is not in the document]
kennygoff: Ok yeah changesets does recommend using a second branch when doing prereleases just so the main branch isn't locked up in prerelease mode. Given we don't make that many changes to this library and we won't be in prerelease for long thats probably still fine to just lock up the main branch but I'd rather not if we don't have to so `main` stays in a releasable state that reflects the latest version.

We can try to run the publish on this branch, we'll need to modify the [GH action](https://github.com/Shopify/worldwide/blob/main/.github/workflows/npm-release.yml) to also run on this branch though so the job that handles publishing will run.

lang/typescript/src/extended-address/splitAddress1.ts
gabypancu: Ah thanks for catching that edge case!

lang/typescript/src/utils/address-fields.ts
gabypancu: Done!

kennygoff (approved):
LGTM 👍 If we want to do the prerelease I still recommend trying to publish the prerelease from this branch instead of merging though as the changesets docs recommend.

lang/typescript/src/utils/address-fields.test.ts
```diff
@@ -214,4 +215,93 @@ describe('splitAddressField', () => {
...
+
+  describe('regexSplitAddressField', () => {
+    test('creates an address object from string matching one of the defined regexes', () => {
+      const fieldDefinition: FieldConcatenationRule[] = [
+        {key: 'streetName'},
+        {key: 'streetNumber'},
+      ];
+      const regexPatterns = [
+        new RegExp('^(?<streetName>[^\\d]+) (?<streetNumber>\\d+)$'),
+      ];
```
DominiqueFlaaa, line 227: nit: can you update this test to have 2 regexPatterns, with the first not a match and the second a match?

.github/workflows/npm-release.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - parse-additional-fields
```
gabypancu, line 7: TODO: remove once done with pre-release

lang/typescript/CHANGELOG.md
```diff
@@ -1,5 +1,31 @@
...
+### Minor Changes
+
+- Add address1 regex for BE, BR, CL, ES, IL, MX
+
+## 0.7.0-next.0
+
+### Minor Changes
+
+- 1f5d405: Add optional tryRegexFallback param to splitAddress1 function to attempt splitting address lines that do not contain the reserved delimiter
```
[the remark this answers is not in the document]
DominiqueFlaaa: >The pre-releases are published and available in [npm](https://www.npmjs.com/package/@shopify/worldwide/v/0.7.0-next.0)

Ah thank you for clarifying.

PR #286, opened 2024-09-24
De address1 regex

gabypancu:

...

...

...

...

### Testing

...

### Checklist


db/data/regions/DE.yml
```diff
@@ -29,6 +29,8 @@ format:
   show: "{firstName} {lastName}_{company}_{address1}_{address2}_{zip} {city}_{country}_{phone}"
 format_extended:
   edit: "{country}_{firstName}{lastName}_{company}_{streetName}{streetNumber}_{address2}_{zip}{city}_{phone}"
+address1_regex:
+- "^(?<streetName>[^\\d,]+[^\\d., ]) *([.,]{1,2})? *(?<streetNumber>\\d+(?: ?[a-z])?)$"
```
gabypancu, line 33: It is quite common to put a period after the street name, sometimes without a space afterward. I also saw a couple instances of a period and a comma 🤷‍♀️

gabypancu: Edit: this is because a period is often used to denote street name abbreviations, so it should not be interpreted as a delimiter and should be left in the street name

rochlefebvre (approved):
Patterns look good! I think this PR's english description of the patterns is outdated.

PR #313, opened 2024-11-28
Add street_number.missing_building_number messages

rochlefebvre:
The street_number field needs message translations for errors related to a possibly missing building number, in instructional and informative formats. The message code and text mirror those for the address1 field.

### Checklist


DominiqueFlaaa:
Please link back to a relevant issue/epic.

rochlefebvre:
> I noticed we have these translations on address2 and line2 as well. Do we really need these?

I don't see a scenario where we would flag address2 or line2 if there was no building number.  Those would go on address1 or street_number.  So I'd say... no?

DominiqueFlaaa (approved):

PR #316, opened 2024-12-11
Speed up configure_i18n by precomputing paths and indexing descendants

gmalette:

In our codebase, we perform `Worldwide.configure_i18n` on every boot, which means every time we run tests. 
Recent versions of Worldwide have decreased the time required from ~350ms down to <200ms, but it's still quite a bit.

Here's a profile of `configure_i18n` taken before this PR:

image

[Profile before](https://github.com/user-attachments/files/18098893/vernier-before.json)

We can see that 2 things take up most of the time: 
- `Dir[]` (`cldr_locale_paths` is clearly visible, 2 others appear nested in `other_data_path` and `regions_data_path`)
- `expanded_locales_from_configuration` via `Array#include?`

Given the files in `data` are static, and the output of `Dir[]` is known at gem-package time, we can use this knowledge to pre-generate files that act as indices to the yaml files.

Second, instead of scanning each locale within the `fallback`, we can make a reverse-index and use that to figure out the descendants. 

image

With these 2 changes, `configure_i18n` goes from 178ms to ~10ms on my machine.

There are other improvements we could make, like performing the matches only on the index (don't need to `match` the `GEM_ROOT` prefix), but I'm happy with this, for now.

Path of least resistance

1. Is there a better place for the rake task? I don't know anything about this gem
2. Are the changes _correct_ ?

### Checklist


rake/tasks/cldr.rake
```diff
@@ -49,6 +49,30 @@ namespace :cldr do
...
     end
 
+    desc <<~DESCRIPTION
+      Generate indices of files in the CLDR data
+
+      This task generates a list of all files in the CLDR data, which is used to speed up the loading of the data.
+
+      eg.: bundle exec rake cldr:data:generate_paths
+    DESCRIPTION
+    task :generate_paths do
```
rochlefebvre, line 59: nit: Can you move the body of this task into its own class, following the pattern of the other tasks in this rake file?
```ruby
# rake/cldr/generate_paths.rb
module Worldwide
  module Cldr
    class GeneratePaths
      def perform
        # ...
```
No need to write tests; none of the other rake helpers do.

data/cldr/paths.txt
rochlefebvre, line 1: I considered the idea of not committing these index files to the repo, and simply having them exist as gem artifacts that get generated during packaging. But this is much better: as index updates will be tracked via commits, their contents will be covered by tests, and there is no chance that a broken packaging workflow will corrupt the indexes.

PR #320, opened 2024-12-16
Add zip requirement and tax inclusive methods to country regions

rochlefebvre:
- Improve test coverage on a few existing country region attributes
- Implement `zip_requirement` on country regions that returns either `required`, `recommended` or `optional` depending on national carrier expectations.
- Implement `tax_inclusive` predicate to indicate if, broadly speaking, the sales tax is included in the price of goods in a region (default `false`).

### Checklist


lib/worldwide/region.rb
```diff
@@ -435,13 +439,23 @@ def zip_autofill
 
     # is a postal code required for this region?
     def zip_required?
-      if zip_requirement.nil?
+      if @zip_requirement.nil?
```
rochlefebvre, line 442: Use direct references to `@zip_requirement` variable to avoid circular referencing between the two methods.

lib/worldwide/region.rb
```diff
@@ -435,13 +439,23 @@ def zip_autofill
...
     end
 
+    # Returns whether (and how firmly) we require a value in the zip field
+    # Possible returns:
+    #  - "required" means a value must be supplied
+    #  - "recommended" means a value is optional, but we recommend providing one
+    #  - "optional" means a value is optional, and we say it is optional
+    # @return [String]
+    def zip_requirement
+      @zip_requirement || (zip_required? ? REQUIRED : OPTIONAL)
```
rochlefebvre, line 456: If `zip_requirement` is not explicitly set on the region yaml, infer requiredness if the region defines a zip regular expression pattern. A region can define a pattern and set zip_requirement to "recommended" or "optional" to avoid this default.

jtrollia: For my knowledge, what are the differences when setting `recommended` VS `optional`?

jtrollia (approved):
✉️

PR #321, opened 2024-12-16
Set Ireland (IE) zip as a required address field

rochlefebvre:
Follows https://github.com/Shopify/worldwide/pull/320

[Since July 2023](https://changelog.shopify.com/posts/shipments-to-the-republic-of-ireland-now-require-postal-code-eircode) Shopify has required postal codes for Ireland addresses. This should be reflected in Worldwide.

### Testing
```
irb(main):002> Worldwide.region(code: "IE").zip_required?
=> true
```

### Checklist

### DIFF
--- a/data/regions/IE.yml
+++ b/data/regions/IE.yml
@@ -13,7 +13,6 @@ tags:
 zip_regex: "^(D6W|[AC-FHKNPRT-Z]\\d{2}) ?[0-9AC-FHKNPRT-Z]{4}$"
 partial_zip_regex: "^(D6W|[AC-FHKNPRT-Z]\\d{2})$"
 zip_example: A65 F4E2
-zip_requirement: recommended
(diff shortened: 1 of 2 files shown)


data/regions/IE.yml
```diff
@@ -13,7 +13,6 @@ tags:
 zip_regex: "^(D6W|[AC-FHKNPRT-Z]\\d{2}) ?[0-9AC-FHKNPRT-Z]{4}$"
 partial_zip_regex: "^(D6W|[AC-FHKNPRT-Z]\\d{2})$"
 zip_example: A65 F4E2
-zip_requirement: recommended
```
rochlefebvre, line 16: As of https://github.com/Shopify/worldwide/pull/320, by removing this field, a country with a zip_regex will lead to `zip_required?` => `true`.

PR #322, opened 2024-12-16
release 1.15.0

rochlefebvre:
Prepare a minor release.

The `Region#zip_requirement` field is no longer nilable, and returns a value even if the country does not set `zip_requirement` in the yaml. Instead of `nil`, it returns `"required"` if `zip_regex` is set or `"optional"` otherwise.  This is a behavior change, but IMO more of a bug fix: a `nil` `zip_requirement` had no defined meaning, and we are clarifying that.

### Checklist


jtrollia (approved):
:shipit:

PR #323, opened 2024-12-18
Set Malaysia and Singapore as tax_inclusive

rochlefebvre:
Shopify considers Malaysia and Singapore to be tax-inclusive countries.

https://www.iras.gov.sg/taxes/goods-services-tax-(gst)/basics-of-gst/invoicing-price-display-and-record-keeping/displaying-and-quoting-prices

I cannot find supporting evidence that prices are indeed tax inclusive in Malaysia 🤷, but for now let's update the yaml to be consistent.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
-Nil.
+- Set Malaysia and Singapore as tax_inclusive [#323](https://github.com/Shopify/worldwide/pull/323)
 
 ---
 
--- a/data/regions/MY.yml
+++ b/data/regions/MY.yml
@@ -2,6 +2,7 @@
 name: Malaysia
 code: MY
 tax: 0.1
+tax_inclusive: true
 currency: MYR
 unit_system: metric
 tax_name: SST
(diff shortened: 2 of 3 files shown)


jtrollia (approved):
Approving with basic questions for my _culture personnelle_ 😅 
- what happens when Shopify and Worldwide disagree on a term?
- what can go wrong when making such changes? Merchants may complain that they start showing prices with taxes included?

rochlefebvre:
> Approving with basic questions for my _culture personnelle_ 😅
> 
> * what happens when Shopify and Worldwide disagree on a term?

Worldwide was created to share Shopify's modeling of countries, zones, languages, and addresses with the world. Most facts contained are held universally true, while some are the result of Shopify picking a value. The `tax_inclusive` field in particular is not especially accurate, as no country is 100% inclusive or exclusive of tax. It's an approximative fallback, really.

So we try to keep worldwide as an accurate modeling of Shopify's world view, rather than the other way around.

> * what can go wrong when making such changes? Merchants may complain that they start showing prices with taxes included?

Worldwide was created as a fork of an internal gem, which still maintains its own set of facts. Our systems currently rely on this internal gem for modeling the world. We are currently reducing this data duplication by having the internal gem call worldwide underneath.

For important yet subjective attributes like `tax_inclusive`, we'd rather maintain the "old" answer, which is that Malaysia and Singapore are mostly tax-inclusive. If I didn't push the present PR out, then our internal gem refactoring would lead to a change in these countries' tax setting.  I want to prevent such a regression by aligning worldwide with the internal gem.

PR #325, opened 2024-12-19
Correct zip regex patterns and make them Javascript-compatible

rochlefebvre:
Update a few broken zip regex patterns and make them compatible with JavaScript.

We will no longer recognize `FPO AE 09397-0007` as a valid US ZIP code. It's not. Only the `09397-0007` part is.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Set Malaysia and Singapore as tax_inclusive [#323](https://github.com/Shopify/worldwide/pull/323)
 - Update or correct example zips [#324](https://github.com/Shopify/worldwide/pull/324)
+- Correct zip regex patterns and make them Javascript-compatible [#325](https://github.com/Shopify/worldwide/pull/325)
 ---
 
 ## [1.15.0] - 2024-12-16


data/regions/MQ.yml
```diff
@@ -13,14 +13,14 @@ group_name: Central America
 tags:
 - EU-OMR
 zip_example: '97220'
-zip_regex: "\\A(MQ-?)?((972)\\d{2,2})\\Z"
+zip_regex: "\\A(MQ-?)?((972)\\d{2,2})\\z"
```
rochlefebvre: The end-of-line sequence is a lowercase z, not Z.

data/regions/US.yml
```diff
@@ -10,7 +10,7 @@ group_name: North America
 name_alternates:
 - United States of America
 zip_label: Zip code
-zip_regex: "(?-mix:\\A\\d{5}(-\\d{4})?\\z)|(?i-mx:\\AFPO\\s*A[A-Z]\\s*\\d{5}-\\d{4}\\z)"
+zip_regex: "(?-mix:\\A\\d{5}(-\\d{4})?\\z)"
```
rochlefebvre: Quoting prior work from @cejaekl:

We noticed that our US regex was defective, and have corrected it. The old regex seems to have been based on a misunderstanding about how the USA Fleet Post Office (FPO) addressing works.

Here's an example FPO address (from https://pe.usps.com/text/pub28/28c2_019.htm):
```
PVT MIKE JONES
PSC 3022 BOX 684
FPO AE 09397-0007
UNITED STATES
```
In this example, the values should be entered into address fields as follows:
 - `name`: `PVT MIKE JONES`
 - `address1`: `PSC 3022 BOX 684`
 - `city`: `FPO` (Fleet Post Office)
 - `province`: `AE` (pseudo-state code for Armed forces, Europe)
 - `zip`: `09397-0007`
 - `country`: `UNITED STATES`

The old regex was permitting values such as
`FPO AE 09397-0007` in the `zip` field.

jtrollia (approved):
Looking good to me 👀

PR #332, opened 2025-02-24
Update known PR zip prefixes and allow zone lookup by partial zip if country hides provinces

rochlefebvre:
We're updating the list of Portugal zip prefixes to be _less_ stale.  The updates in this PR are accurate as of Februrary 2023: still stale, but better.

Also, some wallet pay providers only provide the first part of an address postal code and we still need to look up info on their hidden zone metadata. We're updating `Region#zone(zip:)` to accept an incomplete zip and resolve the first zone that contains this zip, as long as the country hides provinces from addresses.

### Checklist


lib/worldwide/region.rb
```diff
@@ -494,6 +499,13 @@ def province_optional?
       province_optional || !@zones&.any?(&:province?)
     end
 
+    # Returns true if this country has zones defined, and has postal code prefix data for the zones
+    def has_zip_prefixes?
```
rochlefebvre, line 503: I am making this method public.

gabypancu (approved):
Looks good!

>Also, some wallet pay providers only provide the first part of an address postal code and we still need to look up info on their hidden zone metadata. 

Just curious, where did you find this info about wallets?

PR #336, opened 2025-02-28
Remove Ruby 3.1 from CI version matrix; add Ruby 3.4

rochlefebvre:
Remove the three year old Ruby 3.1 from the test matrix and add 3.4. I chose to leave Ruby 3.1 as the gem's minimal supported version.

### Checklist

  - There is no direct impact on the gem. Just what we test thre codebase against.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
       matrix:
         gemfile:
          - Gemfile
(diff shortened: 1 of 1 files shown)


DominiqueFlaaa (approved):
👍 https://github.com/Shopify/worldwide/blob/59cf65a2bc975d722c138eba24c8544b9b022dfe/worldwide.gemspec#L27

PR #339, opened 2025-03-20
Update KwaZulu-Natal province code

LauraAubin:

Updates the province code as requested in https://community.shopify.dev/t/re-shopify-using-the-wrong-iso-code-for-a-province-in-my-country/11663

- Did I miss any other files?
- Is this change accurate?

### DIFF
--- a/data/regions/ZA.yml
+++ b/data/regions/ZA.yml
@@ -49,9 +49,7 @@ zones:
   tax: 0.0
   tax_name: VAT
 - name: KwaZulu-Natal
-  code: NL
-  code_alternates:
-  - KZN
+  code: KZN
   iso_code: ZA-KZN
   tax: 0.0
   tax_name: VAT


gabypancu:
Re-posting my message from [Slack](https://shopify.slack.com/archives/C01M8T03N9W/p1742504527334629?thread_ts=1742498427.199449&cid=C01M8T03N9W) in case anyone else from the team reviews while I'm away
> my concern is that if we remove the NL code, something may break with respect to historical addresses (ex. existing addresses in our DBs). One example is a logged in customer with a saved address in KwaZulu-Natal with province code code NL - when they try to check out in the future using their saved address after we've changed the code, I think they'd see an error saying their province is not valid.  To be safe, we could keep NL as the alternate code.

DominiqueFlaaa:
As discussed internally, we need to do a bit more investigation before committing this change. 

Appreciate you flagging this inconsistency to our team 🙌

PR #342, opened 2025-03-28
[TS] append delimiter to the returned value of `splitAddress2` when missing

jtrollia:

Append the delimiter character to the value returned by `splitAddress2`, when no delimiter was found in the string to split.

This ensures that the whole string is still used as the value of the `line2` field in Checkout's address forms, while offering something to split and concatenate on.

Could this have side-effects on different clients/consumers?

TBC -- should be none

### Testing

This is tricky to tophat as it involves several repositories.

### Checklist

### DIFF
--- a/lang/typescript/src/utils/address-fields.ts
+++ b/lang/typescript/src/utils/address-fields.ts
@@ -45,6 +45,11 @@ export function splitAddressField(
   fieldDefinition: FieldConcatenationRule[],
   concatenatedAddress: string,
 ): Partial<Address> {
+  // Return empty object if no delimiter is found
+  if (!concatenatedAddress.includes(RESERVED_DELIMITER)) {
+    return {};
+  }
+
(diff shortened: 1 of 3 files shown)


lang/typescript/src/utils/address-fields.ts
```diff
@@ -45,6 +45,18 @@ export function splitAddressField(
   fieldDefinition: FieldConcatenationRule[],
   concatenatedAddress: string,
 ): Partial<Address> {
+  if (concatenatedAddress === '') {
+    return {};
+  }
```
jtrollia, line 50: This logic already existed but gets ignore with my new check below.

DominiqueFlaaa (commented):
Clean, well-scoped change 👌 

I think we are missing a changelog entry:  https://github.com/Shopify/worldwide/tree/main/lang/typescript#adding-a-changelog

PR #355, opened 2025-05-30
add HR-CH as neighboring_zones for IN

rrocxela:

https://shopify.slack.com/archives/C06D8A58TPV/p1748607150213969
...

Add the neighboring_zones between CH and HR in India
...

Is this the right solution, to add what we are told by merchants should be considered neighboring zones.
...

CH and HR will be considered neighboring zones, and local delivery will work across those zones.
...

### Testing

...

### Checklist

### DIFF
--- a/data/regions/IN.yml
+++ b/data/regions/IN.yml
@@ -115,6 +115,7 @@ zones:
   tax_name: IGST
   neighboring_zones:
   - PB
+  - HR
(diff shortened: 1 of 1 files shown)


ZhengZhang98 (approved):

gabypancu (approved):
These zones do border each other, so this change looks good to me 👍

PR #357, opened 2025-06-05
Introduce name_alternates for NZ regions, cut v1.17.1

DominiqueFlaaa:
Introduce name_alternates for NZ regions commonly returned by google's APIs. 

Cut new minor version 1.17.1

Introduced name alternates for New Zealand's 16 regions and 1 territory. 

...

Will improve internal systems autocomplete results for NZ. 

### Testing
Tested with the internal system running on this change. Were able to map the google results back to the desired zone code. 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,9 +26,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Security in case of vulnerabilities.
 
 ## [Unreleased]
-- Require postcodes and update regex in MT [#354](https://github.com/Shopify/worldwide/pull/354)
 
 ---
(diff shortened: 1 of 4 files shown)


DominiqueFlaaa:
Tested on our internal service running these changes; [results here](https://github.com/shop/issues-addresses/issues/242#issuecomment-2944876532)

rochlefebvre (commented):
The content changes look good! I recommend we publish this as a patch version increase, not a minor. We're not introducing a new feature here.

In fact, if you can hold off just a little bit, we could also include #354 in the same release. If not, those changes can go in a follow-up patch.

rochlefebvre (approved):

PR #371, opened 2025-06-25
Japanese prefix validation

izi-on:

...

...

...

...

### Testing

...

### Checklist


najibna (commented):
This only adds Hiroshima (`JP-34`) prefixes. Prefer #376 if that lands the full JP set.

PR #376, opened 2025-07-03
Include postal code prefixes for japan zones

izi-on:

We are trying to implement stricter validation for Japanese postal codes.

...

I found a highly credible dataset from [GeoPostcodes](https://portal.geopostcodes.com/).
Then, I created a rake task that would read a CSV from this dataset and populate the existing JP.yml with valid zone prefixes.

- 1. Create a hash with zone codes as the key and an array of their postal codes as a value
- 2. Extract the smallest postal code prefix that will uniquely identify some corresponding zone.
-  -  In order to achieve this 2, I created a trie and populated it with postal codes. When populating, I also marked on the trie node the zone which we are adding the prefix for.  Then, I did a simple DFS traversal and stopped when an encountered node has only one single zone associated with it: this means the current prefix uniquely identifies a zone.
3. Populate the JP yaml with this data

We can now perform naive postal code validation based off purely the zone.

### Testing

For Shopify reviewers, you can compare the results to that of another rake task in `country-db`:

simply run
`rake "postal:codes:import[JP]"`
in country-db and compare the modified JP.yml to this

You will notice that this dataset is more complete

Potential question you might ask: why are there so many postal code prefixes for some zones?
That is because the dataset is such, and I believe this is entirely valid. Let's take an example:

Look at the zip prefixes for JP-02:
'03' and '018-550'

'018-550' seems suspicious, let's investigate. If you open the dataset I linked above, search for "018-55" and notice you will get postal codes in zone JP-02 and JP-05 (018-5511 more specifically). Putting these into google maps:

image

and notice that they are also correspondingly in JP-02 (Aomori Prefecture) and JP-05 (Akita Prefecture):

image

I would like for reviewers **_to test it this out themselves, please pick 5 zip prefixes that seem too long, and test it out yourselves following similar steps as me as written above_**

### Validation

1. I got a sample of 10k real Japanese addresses (For Shopifolk, I linked the methodology for this in the issue linked to this PR)

2. Created a jupyter notebook (`jp-prefix-validator.ipynb`) that will read the CSV, and validate the zip codes against the zip prefixes in JP.yml 

3. Compiled results (For Shopifolk, I reported the results of this validation in the issue linked to this PR) 

### Checklist

### DIFF
--- a/rake/tasks/cldr.rake
+++ b/rake/tasks/cldr.rake
@@ -2,6 +2,8 @@
 
 require_relative "../cldr/locale_generator"
 require_relative "../cldr/patch"
+require_relative "../cldr/region_patcher"
+require_relative "../cldr/csv_reader"
 require_relative "../cldr/puller"
 require_relative "../cldr/cleaner"
 require_relative "../paths/generator"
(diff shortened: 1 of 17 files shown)


rake/tasks/cldr.rake
```diff
@@ -72,5 +74,32 @@ namespace :cldr do
...
       generator.perform
     end
+
+    desc <<~DESCRIPTION
+      Patches Japanese region data with zip code prefixes from a CSV file.
+
+      This task reads a CSV file with Japanese address data, extracts postal
+      code prefixes for each prefecture, and updates the region data files.
+
+      eg.: bundle exec rake "cldr:data:patch_jp_zip_prefixes[path/to/your/file.csv]"
```
izi-on, line 84:
```suggestion
      eg.: bundle exec rake "cldr:data:patch_jp_zip_prefixes[path/to/your/file.csv, source/type]"
```

najibna (commented):
Filling JP `zip_prefixes` is useful. Check the GeoPostcodes license before vendoring the prefixes, and this looks like it would supersede the partial Hiroshima-only #371.

PR #410, opened 2025-12-17
Add ICU4X data generation from worldwide's CLDR

ianks:
This adds Rust infrastructure to generate ICU4X data blobs from worldwide's patched CLDR data. The goal is to enable high-performance i18n operations in native code while ensuring parity with Ruby.

## What's included

- **worldwide-icu4x-data**: Crate that embeds the generated ICU4X postcard blob and provides a buffer provider
- **worldwide-icu4x-datagen**: Converts worldwide's YAML data to CLDR JSON format, then runs ICU4X datagen
- **worldwide-icu4x-parity**: CLI for parity testing between Ruby and ICU4X implementations
- **Parity tests**: Exhaustive tests comparing Ruby and ICU4X output for all locales

## Currently supported

- Plural rules (cardinal and ordinal) - 170+ locales
- List formatting (and/or/unit patterns)

## Not yet supported

Decimal formatting is blocked because ICU4X requires `decimalFormats-numberSystem-latn` with format patterns (e.g., `#,##0.###`) in addition to symbols. Worldwide's YAML only has symbols. The parity tests are in place but skipped.

## Usage

```bash
# Generate the ICU4X blob
bundle exec rake icu4x:generate

# Run parity tests
bundle exec rake icu4x:test
```


najibna (commented):
ICU4X datagen plus the Ruby parity tests are a solid direction. The postcard blob is a large generated artifact — keep regeneration (`icu4x:generate`) documented so reviewers do not have to inspect the binary.

PR #462, opened 2026-05-07
Update BR regex to split address1 properly

jtrollia:

Today, BR's `splitAddress1` regex only matches 2-component canonical input (`<street>`, `<number>`). When given 3-component input (`<street>`, `<number>`, `<complement>` — common in autofill payloads from Chrome, password managers, and PayPal), the regex bails to the fallback and returns the entire string in `streetName`.

This adds an optional third capture group for line2 so 3-component inputs are parsed correctly. The change is in BR's region data only; no library code changes.

Backward compatibility: existing 2-component inputs continue to match the regex with no `line2` field — confirmed via existing test cases (all unchanged). Adding `line2` to the result is additive — consumers reading `streetName`/`streetNumber` are unaffected; consumers wanting `line2` can opt in.

Note: this leaves `combined_address_format` untouched. `line2` remains canonically in `address2`. The regex is for parsing flat natural-language strings (which mix all components); the canonical format is for structured serialization. The library already supports this asymmetry by design.

...

...

...

### Testing

...

### Checklist

### DIFF
--- a/data/regions/BR.yml
+++ b/data/regions/BR.yml
@@ -29,7 +29,7 @@ format:
 format_extended:
   edit: "{country}_{firstName}{lastName}_{company}_{zip}_{streetName}{streetNumber}_{line2}{neighborhood}_{city}{province}_{phone}"
 address1_regex:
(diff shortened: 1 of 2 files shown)


jtrollia:
Closing in favour of https://github.com/Shopify/worldwide/pull/520

PR #465, opened 2026-05-08
dependabot: remove unnecessary registry to fix failing update jobs

DominiqueFlaaa:
## Problem

[Dependabot update jobs are failing ](https://github.com/Shopify/worldwide/actions/runs/25574807674/job/75079245836)at startup before any dependency resolution:

```
Error: fetching credentials: unexpected status code: 422:
  {"errors":[{"status":422,"title":"Secret Not Found","detail":"DEPENDENCIES_GITHUB_USER"}]}
```

The `dependabot.yml` configures a `github-com` git registry using `DEPENDENCIES_GITHUB_USER` / `DEPENDENCIES_GITHUB_TOKEN` secrets that don't exist in this repo. Dependabot fetches credentials for all configured registries at job startup, so every job fails — including `npm` jobs in `lang/typescript` that have nothing to do with GitHub-hosted gems.

## Why the registry is unnecessary

Dependabot accesses public GitHub repos via its own built-in token; no explicit `git` registry entry is needed.  Default behavior is to raise pull requests only to update dependencies stored in publicly accessible registries.

`registries` are only necessary [when accessing private package registries](https://docs.github.com/en/enterprise-cloud@latest/code-security/reference/supply-chain-security/dependabot-options-reference#registries--). Worldwide does not leverage any private packages, and so this config is not needed. 

**Precedent:** `Shopify/ruby-lsp` started with the exact same boilerplate (commit [`90afd3e`](https://github.com/Shopify/ruby-lsp/commit/90afd3e273)). Five days later, ["Simplified dependabot config"](https://github.com/Shopify/ruby-lsp/commit/9d708c826f) removed it. Their Dependabot runs have been clean ever since.

## Changes

- **Remove** all `registries`
- **Add** `npm` ecosystem for `lang/typescript` on a daily schedule — Dependabot was already auto-generating jobs for this directory; making it explicit gives us control

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,15 +1,4 @@
 version: 2
-registries:
-  ruby-shopify:
-    type: rubygems-server
-    url: https://pkgs.shopify.io/basic/gems/ruby
-    username: ${{secrets.RUBYGEMS_SERVER_PKGS_SHOPIFY_IO_USERNAME}}
-    password: ${{secrets.RUBYGEMS_SERVER_PKGS_SHOPIFY_IO_PASSWORD}}
-  github-com:
-    type: git
-    url: https://github.com
-    username: ${{secrets.DEPENDENCIES_GITHUB_USER}}
-    password: ${{secrets.DEPENDENCIES_GITHUB_TOKEN}}
 updates:
   - package-ecosystem: bundler
     directory: "/"
@@ -17,7 +6,10 @@ updates:
       interval: daily
     open-pull-requests-limit: 100
     insecure-external-code-execution: allow
-    registries: "*"
+  - package-ecosystem: npm
+    directory: "/lang/typescript"
+    schedule:
+      interval: daily
   - package-ecosystem: github-actions
     directory: "/"
     schedule:


.github/dependabot.yml
```diff
@@ -5,19 +5,19 @@ registries:
     url: https://pkgs.shopify.io/basic/gems/ruby
     username: ${{secrets.RUBYGEMS_SERVER_PKGS_SHOPIFY_IO_USERNAME}}
     password: ${{secrets.RUBYGEMS_SERVER_PKGS_SHOPIFY_IO_PASSWORD}}
-  github-com:
-    type: git
-    url: https://github.com
-    username: ${{secrets.DEPENDENCIES_GITHUB_USER}}
-    password: ${{secrets.DEPENDENCIES_GITHUB_TOKEN}}
```
AndreaSmith053122, line 12: This github registry has been here since [2024](https://github.com/Shopify/worldwide/commit/b13b07f399cf0a56dbea716d76ccfe60f1d9aefa) and there are no recent changes to this file. Why did it suddenly become a problem?

.github/dependabot.yml
```diff
@@ -1,23 +1,15 @@
 version: 2
-registries:
-  ruby-shopify:
```
DominiqueFlaaa, line 3: `ruby-shopify` registry is only relevant if a gem in the dependency graph is sourced from pkgs.shopify.io

.github/dependabot.yml
DominiqueFlaaa: The action has been failing since the start -- https://github.com/Shopify/worldwide/actions/runs/14199318446

I assume it was not previously part of the shipit pipeline as a required step. Worldwide uses the default config...let me dig 🧐

DominiqueFlaaa: I dont have a clear answer why these failures are gating deploys. Considering they have been failing for over a year, it should be safe to clean up the registries.

AndreaSmith053122 (approved):

DominiqueFlaaa:
[Run](https://github.com/Shopify/worldwide/runs/75093368898) was a success! 🎉 

Screenshot 2026-05-08 at 4 56 27 PM

PR #485, opened 2026-05-11
Pin minimatch via pnpm overrides to resolve security advisory

DominiqueFlaaa:
## Why

`minimatch@3.1.2` and `minimatch@9.0.4` are flagged by a GitHub security advisory. Dependabot raised a security update job to patch this, but it failed with:

```
Dependabot::NpmAndYarn::FileUpdater::NoChangeError
No files were updated! Package manager: pnpm
```

Failing build: https://github.com/Shopify/worldwide/actions/runs/25670184913/job/75353085843

The advisory affected version ranges from the job log:
- `< 3.1.3` (affects `3.1.2`)
- `>= 9.0.0 < 9.0.7` (affects `9.0.4`)

## Root cause

Same pattern as #484 (postcss). `minimatch` is not a direct dependency — it is pulled in transitively. With pnpm v9, `pnpm update <pkg>` only touches packages declared in `package.json`, so Dependabot's update command produces no lockfile diff and aborts.

## Resolution

Added version-scoped `pnpm.overrides` entries to force safe resolutions within each major version:

```json
"pnpm": {
  "overrides": {
    "minimatch@3": "^3.1.3",
    "minimatch@9": "^9.0.7"
  }
}
```

This resolves `minimatch@3.1.2 → 3.1.5` and `minimatch@9.0.4 → 9.0.9` in the lockfile, both above the patched thresholds.

The version-scoped selector (`@3`, `@9`) is important — a bare `minimatch: ">=3.1.3"` override would let pnpm resolve all the way to the latest major (10.x), which would be a breaking change for packages depending on v3 semantics.

### DIFF
--- a/lang/typescript/package.json
+++ b/lang/typescript/package.json
@@ -70,7 +70,9 @@
   "prettier": "@shopify/prettier-config",
   "pnpm": {
     "overrides": {
-      "postcss": ">=8.5.10"
+      "postcss": ">=8.5.10",
+      "minimatch@3": "^3.1.3",
+      "minimatch@9": "^9.0.7"
     }
   }
 }
--- a/lang/typescript/pnpm-lock.yaml
+++ b/lang/typescript/pnpm-lock.yaml
@@ -6,6 +6,8 @@ settings:
 
 overrides:
   postcss: '>=8.5.10'
+  minimatch@3: ^3.1.3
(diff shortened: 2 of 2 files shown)


DominiqueFlaaa:
closed in favour of https://github.com/Shopify/worldwide/pull/486

PR #493, opened 2026-05-18
Remove optional ? in PT zip_regex

AndreaSmith053122:

Portugal postal code format should be XXXX-XXX, all numeric. Our current regex does not require the second set of three digits, despite it being the required format since 1998.

Contributes to https://github.com/shop/issues-checkout/issues/11587

- Remove the `?` part of the regex making the second set of digits optional.
- Leave the `?` that makes the `-` optional.

I don't think there's any way to flag these changes for a gradual rollout, but I'd be happy to hear otherwise.

[After services start using the new gem version]
- Customers will be required to enter the full postal code
- Merchants touching old records will need to put in the full postal code. We expect Atlas to be able to provide the correct value.

### Checklist

### DIFF
--- a/data/regions/PT.yml
+++ b/data/regions/PT.yml
@@ -8,7 +8,7 @@ tax_name: VAT
 tax_inclusive: true
 group: European Countries
 group_name: Europe
-zip_regex: "^(PT?-?)?\\d{4}(-?\\d{3})?$"
+zip_regex: "^(PT?-?)?\\d{4}-?\\d{3}$"
 zip_example: 2725-079
 phone_number_prefix: 351
 building_number_required: true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ---
 
-## [1.24.3] - 2026-05-18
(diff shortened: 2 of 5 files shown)


ananyaneogi:
Might be good to add a [changelog post](https://changelog.shopify.com/admin/posts)  ? Given that anyone consuming worldwide will start rejecting 4-digit PT zips now.

data/regions/PT.yml
```diff
@@ -8,7 +8,7 @@ tax_name: VAT
 tax_inclusive: true
 group: European Countries
 group_name: Europe
-zip_regex: "^(PT?-?)?\\d{4}(-?\\d{3})?$"
```
ananyaneogi, line 11: lets add a test for this regex behaviour change

AndreaSmith053122:
> Might be good to add a [changelog post](https://changelog.shopify.com/admin/posts) ? Given that anyone consuming worldwide will start rejecting 4-digit PT zips now.

@ananyaneogi I reached out to @dennishcodes to create the changelog post.

dennishcodes:
draft changelog [here](https://changelog.shopify.com/admin/posts/checkout-now-requires-full-portuguese-postal-codes/edit), i put a very far out publication date so we can adjust it when the right date is known

ananyaneogi (approved):

PR #520, opened 2026-07-07
Tighten BR address1 regex fallback

jtrollia:
## Summary

This consolidates the BR address1 parser changes so Worldwide can handle both:

- comma-delimited raw prefill values whose street names can contain digits, like `Rua 25 de Marco, 100`
- flat 3-component values that include complement text, like `Praca Ramos de Azevedo, 123, Apto 45`

The BR regex list is intentionally ordered from higher-confidence comma-delimited patterns to the existing digit-free street-name fallbacks. Ambiguous values such as `Avenida Normando Tedesco, 1400 - Centro` continue to fall back to the full `streetName` instead of being split heuristically.

## Follow-up

Checkout-web still needs to bump `@shopify/worldwide` and decide how to seed a returned `line2` into checkout state at the trusted prefill/autofill boundary.

## Validation

- `pnpm test src/extended-address/splitAddress1.test.ts`

### DIFF
--- a/data/regions/BR.yml
+++ b/data/regions/BR.yml
@@ -29,6 +29,10 @@ format:
 format_extended:
   edit: "{country}_{firstName}{lastName}_{company}_{zip}_{streetName}{streetNumber}_{line2}{neighborhood}_{city}{province}_{phone}"
 address1_regex:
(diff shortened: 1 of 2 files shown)


rochlefebvre (approved):

oliverigor (approved):
:lets-go:

PR #527, opened 2026-07-08
Use OIDC for npm publishing

jtrollia:
## Summary

Updates the TypeScript package release workflow to publish through npm trusted publishing/OIDC instead of the legacy secret-backed npm token.

This follows the standard OIDC path by:

- running the release on Node 24
- using changesets/action v1.9.0, which avoids writing a token auth line when NPM_TOKEN is empty
- keeping NPM_TOKEN as the required literal empty string
- keeping provenance enabled
- routing the actual package upload through npm CLI 11.5.1 while preserving the already-published no-op guard

Once the trusted publisher migration is active, this should allow @shopify/worldwide to publish through OIDC.

### DIFF
--- a/.github/workflows/npm-release.yml
+++ b/.github/workflows/npm-release.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6.2.0
         with:
-          node-version: 20
+          node-version: 24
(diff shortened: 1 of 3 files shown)


oliverigor (approved):
🤞🏽

rochlefebvre (approved):
Seems to align with the example workflow definition on the Vault page. We'll see if it works after it merges.

PR #529, opened 2026-07-13
Fix Nova Scotia HST default rate

xuyuanx:

Nova Scotia reduced its HST rate from 15% to 14% effective April 1, 2025. Worldwide still exposes Nova Scotia's `tax_rate` as `0.15`, which causes consumers of Worldwide region data to keep using the old default.

Related issue: https://github.com/shop/issues-taxes/issues/3443

Following the same pattern as https://github.com/Shopify/worldwide/pull/425: update the static region data and add a changelog entry under Unreleased. I also added a regression assertion for `Worldwide.region(code: "CA").zone(code: "NS").tax_rate` so this doesn't drift back to 15%.

Is `data/regions/CA.yml` the right data source for the province-level default tax rate, and should any downstream consumers be explicitly notified to bump Worldwide after this lands/releases?

Consumers reading `Worldwide.region(code: "CA").zone(code: "NS").tax_rate` will get `0.14` instead of `0.15` once they consume a Worldwide release containing this change.

### Testing

- `ruby -Ilib -e 'require "worldwide"; puts Worldwide::VERSION; puts Worldwide.region(code: "CA").zone(code: "NS").tax_rate'` → prints `1.25.4` and `0.14`
- Attempted `bundle exec ruby -Itest test/worldwide/region_test.rb`, but local bundle setup is blocked by native extension build failures in this environment (`bigdecimal`/`byebug` missing compiler/jemalloc headers). CI should run the full test suite.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Security in case of vulnerabilities.
 
 ## [Unreleased]
-- Fix Nova Scotia HST rate from 15% to 14%
+- Fix Nova Scotia HST rate from 15% to 14% [#529](https://github.com/Shopify/worldwide/pull/529)
 
 ---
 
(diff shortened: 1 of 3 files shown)


gabypancu:
Can you link your PR in the changelog 🙂 You can look at the other changelog entries for reference

xuyuanx:
Done in 692d79f9 — the changelog entry now links #529.

gabypancu (approved):

PR #540, opened 2026-07-21
Bump i18n from 1.14.1 to 1.15.2 and fix Cldr fiber-local config

trishrempel:
## What

Bumps `i18n` from 1.14.1 to 1.15.2 (superseding dependabot #509) and fixes the test regression that bump introduced in `Worldwide::Cldr`, adds a matching gemspec constraint, and adds regression tests.

## Why

i18n 1.15.0 made both the active config **and** the fallbacks fiber-local: on Ruby 3.2+ they are stored in `Fiber[:i18n_config]` / `Fiber[:i18n_fallbacks]` (with thread-local storage as the fallback when fibers do not support `[]`, e.g. Ruby 3.1). See ruby-i18n/i18n#729 and #731.

`Worldwide::Cldr#with_cldr` set the CLDR config and fallbacks by writing `Thread.current[:i18n_config]` and `Thread.current[:i18n_fallbacks]` directly. Under 1.15 those assignments are no longer read on Ruby 3.2+, so the CLDR config (and its custom exception handler) and the CLDR fallbacks were never applied. That regressed currency, number, date, and weekday name formatting: `bundle exec rake test` went from all green to 8 failures + 1 error (e.g. `€1,234.50` rendered as `EUR1,234.50`, `Kč` came back `nil`, `zh-HK` number formatting fell back, stand-alone weekday/month names broke).

## Fix

`with_cldr` now saves, swaps, and restores the config and fallbacks through a storage strategy that mirrors i18n 1.15's own storage. Because i18n keeps config and fallbacks in Fiber storage on Ruby 3.2+ and in thread-local storage on Ruby 3.1 (i18n 1.15 still supports Ruby 3.1), the choice is captured **once** at load time:

- `STORAGE = Fiber.respond_to?(:[]) ? FiberStorage : ThreadStorage` selects the strategy once, so there is no per-call reflection or repeated conditional on the hot path.
- `FiberStorage` reads/writes `Fiber[:i18n_config]` (setting `owner` to the current fiber, matching `I18n.config=`) and `Fiber[:i18n_fallbacks]`.
- `ThreadStorage` reads/writes `Thread.current.thread_variable_*(:i18n_config)` and `Thread.current[:i18n_fallbacks]`, matching i18n's thread-local layout exactly.

The fallbacks slot is set directly rather than through `I18n.fallbacks=` on purpose: `I18n.fallbacks=` also mutates the shared class-level `@@fallbacks`, whereas we keep the swap fiber/thread-local so it never clobbers the host application's global fallbacks. Because Fiber storage is per-root-fiber, this is also thread-local, so concurrent callers never see each other's CLDR config or fallbacks.

`worldwide.gemspec` now pins `i18n >= 1.15` so the storage assumption is enforced for consumers (the gem previously declared `i18n` with no constraint).

This is the same class of fix as Shopify/shopify-i18n#2273, extended to also cover the fiber-aware fallbacks storage that i18n 1.15 introduced (1.14.8, which that PR targeted, only moved the config to a thread variable).

## Testing

- New `test/worldwide/cldr_test.rb` covers: CLDR config/fallbacks actually apply while translating (the regression guard, `€1,234.50 EUR`), config + fallbacks are restored after a normal return and after the translation raises, and the host application's custom `I18n.fallbacks` is preserved across a Cldr translation.
- `bundle exec rake test` passes (1397 tests, 0 failures, 0 errors) and `rubocop` is clean. Before the fix the same suite reported 8 failures + 1 error, and reverting to i18n 1.14.1 is all green, confirming the regression is caused by the bump.
- Thread safety verified out of band: 20 threads × 500 iterations of `Worldwide.currency(...).format_explicit(...)`, each with a distinct host `I18n.fallbacks`, produced correct output in every thread with no fallbacks leakage.

Co-authored-by: Claude Opus 4.8 <noreply@anthropic.com>
Orchestrated-by: ae <noreply@shopify.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Security in case of vulnerabilities.
 
 ## [Unreleased]
+- Bump i18n from 1.14.1 to 1.15.2 (now required as `>= 1.15`) and update `Cldr.with_cldr` to mirror i18n's fiber-aware config and fallbacks storage so the CLDR config still applies on Ruby 3.2+. [#540](https://github.com/Shopify/worldwide/pull/540)
 
 ---
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -15,7 +15,7 @@ PATH
   specs:
     worldwide (1.25.5)
       activesupport (>= 7.0)
-      i18n
+      i18n (>= 1.15)
       phonelib (~> 0.8)
 
 GEM
@@ -42,7 +42,7 @@ GEM
     connection_pool (2.4.1)
     drb (2.1.1)
       ruby2_keywords
-    i18n (1.14.1)
+    i18n (1.15.2)
       concurrent-ruby (~> 1.0)
     json (2.7.1)
     language_server-protocol (3.17.0.6)
--- a/lib/worldwide/cldr.rb
+++ b/lib/worldwide/cldr.rb
@@ -11,6 +11,54 @@ module Worldwide
       cldr_config.exception_handler = Worldwide::Config.exception_handler
     end
 
+    # i18n 1.15 keeps the active config and fallbacks in Fiber storage on Ruby
+    # 3.2+ and in thread-local storage otherwise. We must write the same slots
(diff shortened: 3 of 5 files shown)


rochlefebvre (approved):
Looks good!

PR #563, opened 2026-07-29
Prepare patch release 1.26.1

SeriouslyAwesome:
Patch release so the IE `zips_crossing_provinces` fix in #562 — the only change since `v1.26.0` — gets published.

- Bump `Worldwide::VERSION` and the lockfile version to 1.26.1
- Move the `[Unreleased]` entry into a dated 1.26.1 section

Publishing and tag creation are left to ShipIt after merge.

### Testing

- `bundle exec rake`: 1444 tests, 0 failures; RuboCop 107 files, 0 offenses
- `bundle exec gem build worldwide.gemspec` → 1.26.1

### Checklist


DominiqueFlaaa (approved):
Image

PR #577, opened 2026-08-21
Parse Brazilian comma-delimited address complements

jtrollia:

Brazilian raw address lines can place a recognized complement directly after a comma-delimited building number, for example:

- `Avenida Felipe Wandscheer, 7335 Condominio Reserva IguaÇu`
- `Praça Ramos de Azevedo, 123 Apto 45`

The current ordered BR fallback patterns do not recognize this shape, so `splitAddress1` leaves the entire value in `streetName` instead of extracting the building number and complement.

Add one high-confidence BR fallback pattern after the existing comma-delimited patterns. It requires:

- a comma before the building number
- a valid existing BR building-number shape
- one of the complement keywords already trusted by the no-comma fallback, or `Condominio` / `Condomínio`
- content following the complement keyword

This follows the conservative ordered-pattern approach from #520 without treating arbitrary text after a number as `line2`.

The boundary of the trusted complement vocabulary. Inputs such as `Avenida Felipe Wandscheer, 7335 Centro`, bare complement keywords, plural `Condominios`, and comma-less numeric street shapes continue to fall back to the full `streetName`.

`splitAddress1("BR", address, true)` now extracts `streetName`, `streetNumber`, and `line2` from comma-delimited numbers followed by `Apto`, `Apartamento`, `Bloco`, `Casa`, `Sala`, `Fundos`, `Loja`, or `Condominio` / `Condomínio` complements.

A patch changeset is included for the TypeScript package.

### DIFF
--- a/data/regions/BR.yml
+++ b/data/regions/BR.yml
@@ -31,6 +31,7 @@ format_extended:
 address1_regex:
   - "^(?<streetName>.+?),\\s*(?<streetNumber>(?:\\d+[a-zà-öø-ÿ]?|\\d+(?:[-/]\\d+[a-zà-öø-ÿ]?)+|\\d+(?:\\s+\\d+)+|s\\/?n|sem n[úu]mero))$"
   - "^(?<streetName>.+?),\\s*(?<streetNumber>(?:\\d+[a-zà-öø-ÿ]?|\\d+(?:[-/]\\d+[a-zà-öø-ÿ]?)+|\\d+(?:\\s+\\d+)+|s\\/?n|sem n[úu]mero))\\s*,\\s*(?<line2>.+)$"
+  - "^(?<streetName>.+?),\\s*(?<streetNumber>(?:\\d+[a-zà-öø-ÿ]?|\\d+(?:[-/]\\d+[a-zà-öø-ÿ]?)+|\\d+(?:\\s+\\d+)+|s\\/?n|sem n[úu]mero))\\s+(?<line2>condom[ií]nio(?:\\s+.+)?)$"
   - "^(?<streetName>(?:[^\\d,\\s]+\\s)*[^\\d,\\s]+)\\s+(?<streetNumber>\\d+(?: ?[a-z])?)\\s+(?<line2>(?:apt(?:o|\\.)?|apartamento|bloco|casa|sala|fundos|loja)\\b.+)$"
   - "^(?<streetName>(?:[^\\d,\\s]+\\s)*[^\\d,\\s]+)(?:,? ?)(?<streetNumber>\\d+(?: ?[a-z])?)\\s*,\\s*(?<line2>.+)$"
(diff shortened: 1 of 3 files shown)


data/regions/BR.yml
```diff
@@ -31,6 +31,7 @@ format_extended:
 address1_regex:
   - "^(?<streetName>.+?),\\s*(?<streetNumber>(?:\\d+[a-zà-öø-ÿ]?|\\d+(?:[-/]\\d+[a-zà-öø-ÿ]?)+|\\d+(?:\\s+\\d+)+|s\\/?n|sem n[úu]mero))$"
   - "^(?<streetName>.+?),\\s*(?<streetNumber>(?:\\d+[a-zà-öø-ÿ]?|\\d+(?:[-/]\\d+[a-zà-öø-ÿ]?)+|\\d+(?:\\s+\\d+)+|s\\/?n|sem n[úu]mero))\\s*,\\s*(?<line2>.+)$"
+  - "^(?<streetName>.+?),\\s*(?<streetNumber>(?:\\d+[a-zà-öø-ÿ]?|\\d+(?:[-/]\\d+[a-zà-öø-ÿ]?)+|\\d+(?:\\s+\\d+)+|s\\/?n|sem n[úu]mero))\\s+(?<line2>condom[ií]nio(?:\\s+.+)?)$"
```
oliverigor: I think this pattern is narrower than it should. 

The shape it fixes,  `streetName, number <keyword complement>`,is equally unhandled for the keyword list we already trust in the no-comma pattern below: 

`Praça Ramos de Azevedo, 123 Apto 45` still falls through everything and returns the full string as `streetName` (I checked against the compiled pattern list). So we'd fix condominio here and then presumably drip in `apto`, `casa`, etc. one PR at a time.

What about reusing the existing keyword alternation as the `line2` anchor instead ?

```
 ^(?<streetName>.+?),\s*(?<streetNumber>(?:\d+[a-zà-öø-ÿ]?|\d+(?:[-/]\d+[a-zà-öø-ÿ]?)+|\d+(?:\s+\d+)+|s\/?n|sem n[úu]mero))\s+(?<line2>(?:apt(?:o|\.)?|apartamento|bloco|casa|sala|fundos|loja|condom[ií]nio)\b.+)$
```

I checked (🤖)  and this parses both of your new positive cases identically, additionally handles Apto 45 / Casa 2 complements, and still falls back on all the ambiguous cases in your tests (7335 Centro, 1400 - Centro, Rua 4 Condominio X, and plural Condominios Verdes). 

As a bonus, `\b.+` also fixes a small inconsistency in the current version: `(?:\s+.+)?` lets Rua X, 100 Condominio parse with `line2: "Condominio"`, a complement that carries no information, the no-comma keyword pattern requires content after the keyword, and I think this one should too. wdyt ?

jtrollia: Good catch — updated in 6c8f7964. The new pattern reuses the existing trusted complement vocabulary plus Condominio / Condomínio. I added comma-delimited Apto and Casa positives, along with guards for bare keywords and plural Condominios; the prior ambiguity guards still pass. The full TypeScript suite is green with 165 tests.

rochlefebvre (approved):

PR #582, opened 2026-08-26
Resolve Gregorian weekday/month names with CLDR inheritance

### LINKED ISSUE #581: Gregorian#month_names and #weekday_names return partial data for 104 locales (en-CA, en-GB, en-IN, ...)
### Summary

`Worldwide::Calendar::Gregorian.month_names` and `.weekday_names` return **partial** results for 104 locales, including `en-CA`, `en-GB`, `en-IN`, `en-AU`, `en-NZ`, `en-ZA` and every other `en-001` descendant.

```ruby
Worldwide::Calendar::Gregorian.month_names(locale: "en-CA", width: :abbreviated)
#=> ["Sept"]
# expected: ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sept", "Oct", "Nov", "Dec"]

Worldwide::Calendar::Gregorian.weekday_names(locale: "se-FI")
#=> {mon: "mánnodat", tue: "disdat", thu: "duorastat", sat: "lávvordat"}
# expected: all seven, with sun/wed/fri inherited from `se`
```

Reproduced on `worldwide` 1.26.2 (current `main`, `ad6eeefb`), `i18n` 1.15.2, Ruby 3.4.8.

### Cause

CLDR resolves locale data **item by item**: the effective data for a locale is the union of its own data with each of its ancestors', with the more specific locale winning per item (UTS #35 §4.1). Locale files in `data/cldr/locales/` are stored as deltas accordingly — `data/cldr/locales/en-CA/calendars.yml` contains exactly:

```yaml
months:
  stand_alone:
    abbreviated:
      9: Sept
```

`I18n::Backend::Fallbacks#translate` returns the **first non-nil node** it finds in the fallback chain. That is correct for leaves (`months.stand_alone.abbreviated.1` still resolves to `"Jan"` from `en`) but wrong for a whole node: `months.stand_alone.abbreviated` matches `en-CA`'s one-entry hash and the eleven inherited entries never get merged in.

`month_names` then calls `.values` on that hash, so the truncation is silent — `lib/worldwide/calendar/gregorian.rb#L20`.

### Affected locales

805 known locales × 2 methods × 3 widths = 4830 combinations; 108 of them are wrong, spanning 104 distinct locales.

| Method | Width | Count | Locales |
|---|---|---|---|
| `month_names` | `:abbreviated` | 101 | `ar-DZ`, `ar-MA`, `ar-TN`, `en-001`, `en-150`, `en-AG`, `en-AI`, `en-AT`, `en-BB`, `en-BE`, `en-BM`, `en-BS`, `en-BW`, `en-BZ`, `en-CA`, `en-CC`, `en-CH`, `en-CK`, `en-CM`, `en-CX`, `en-CY`, `en-DE`, `en-DG`, `en-DK`, `en-DM`, `en-ER`, `en-FI`, `en-FJ`, `en-FK`, `en-FM`, `en-GD`, `en-GG`, `en-GH`, `en-GI`, `en-GM`, `en-GY`, `en-HK`, `en-IE`, `en-IL`, `en-IM`, `en-IN`, `en-IO`, `en-JE`, `en-JM`, `en-KE`, `en-KI`, `en-KN`, `en-KY`, `en-LC`, `en-LR`, `en-LS`, `en-MG`, `en-MO`, `en-MS`, `en-MT`, `en-MU`, `en-MV`, `en-MW`, `en-MY`, `en-NA`, `en-NF`, `en-NG`, `en-NL`, `en-NR`, `en-NU`, `en-NZ`, `en-PG`, `en-PK`, `en-PN`, `en-PW`, `en-RW`, `en-SB`, `en-SC`, `en-SD`, `en-SE`, `en-SG`, `en-SH`, `en-SI`, `en-SL`, `en-SS`, `en-SX`, `en-SZ`, `en-TC`, `en-TK`, `en-TO`, `en-TT`, `en-TV`, `en-TZ`, `en-UG`, `en-VC`, `en-VG`, `en-VU`, `en-WS`, `en-ZA`, `en-ZM`, `en-ZW`, `hi-Latn`, `hi-Latn-IN`, `se-FI`, `sr-Latn-ME`, `sr-Latn-XK` |
| `month_names` | `:narrow` | 3 | `ar-DZ`, `ar-MA`, `ar-TN` |
| `month_names` | `:wide` | 3 | `ar-DZ`, `ar-MA`, `ar-TN` |
| `weekday_names` | `:wide` | 1 | `se-FI` |

Worst cases return a single entry: `hi-Latn` and `se-FI` return 1 of 12 abbreviated month names; `sr-Latn-ME` and `sr-Latn-XK` return 5 of 12. `en-GB` and friends return 1 of 12.

### Secondary issue: missing structural keys degrade into a plausible string

`Worldwide::I18nExceptionHandler#degraded_translation` (`lib/worldwide/i18n_exception_handler.rb#L51-L57`) turns any unresolved key into a humanized copy of its last segment. For a structural lookup that means a missing weekday table comes back as the **string** `"wide"` rather than raising:

```ruby
Worldwide::Cldr.t("calendars.gregorian.days.stand_alone.wide", locale: :some_locale_without_data)
#=> "wide"
```

That is a reasonable default for leaf strings in a UI, but for a hash-valued CLDR node it converts a data gap into a plausible-looking value, and `month_names` then fails with `NoMethodError: undefined method 'values' for an instance of String` some distance from the cause. (This is how the same call sites failed on worldwide < 1.25.6, before the fiber-storage fix in 76762d19 restored the CLDR fallback chain on Ruby 3.2+.)

### Proposed fix

Resolve structural CLDR keys with CLDR's inheritance rules — walk the fallback chain from the least specific ancestor to the most specific, deep-merging each locale's own contribution, and resolve `<alias>` nodes against the *requested* locale rather than the ancestor they were found in. Then have `weekday_names` / `month_names` raise if an entry is still missing rather than returning degraded data.

PR: #NNN

nightsurgex2:

Fixes #581.

`Worldwide::Calendar::Gregorian.month_names` and `.weekday_names` return **partial** results for 104 locales, including `en-CA`, `en-GB`, `en-IN`, `en-AU`, `en-NZ`, `en-ZA` and every other `en-001` descendant:

```ruby
Worldwide::Calendar::Gregorian.month_names(locale: "en-CA", width: :abbreviated)
#=> ["Sept"]                     # before
#=> ["Jan", ..., "Sept", ...]    # after, all twelve

Worldwide::Calendar::Gregorian.weekday_names(locale: "se-FI")
#=> {mon: "mánnodat", tue: "disdat", thu: "duorastat", sat: "lávvordat"}   # before
#=> all seven, sun/wed/fri inherited from `se`                             # after
```

CLDR resolves locale data item by item: a locale's effective data is the union of its own data with each of its ancestors', the more specific locale winning per item (UTS #35 §4.1). The YAML in `data/cldr/locales/` is stored as deltas to match — `en-CA/calendars.yml` contains only `months.stand_alone.abbreviated: {9: Sept}`.

`I18n::Backend::Fallbacks#translate` returns the first non-nil node in the chain. Correct for leaves (`...abbreviated.1` still resolves to `"Jan"` from `en`), wrong for a whole node: the lookup stops at `en-CA`'s one-entry hash and the eleven inherited entries are never merged in. `month_names` then calls `.values` on it, so the truncation is silent.

Full affected-locale table is in the issue. 108 of 4830 method × width × locale combinations were wrong.

**Chosen: resolve structural keys with CLDR's inheritance rules at read time.** `Worldwide::Cldr.resolved_hash` walks the fallback chain from the least specific ancestor to the most specific, deep-merging each locale's own contribution, and resolves CLDR `<alias>` nodes against the *requested* locale rather than the ancestor they were found in. That is what UTS #35 specifies, it is a handful of lines, and it leaves `t` — and therefore every leaf lookup in the gem — untouched.

Rejected alternatives:

- **Flatten the inheritance at generation time** (`rake/cldr/locale_generator.rb`), writing fully-resolved calendar data into each locale file. Correct output, but it discards the delta structure the CLDR import produces and multiplies `data/cldr/locales/` for 805 locales. It also has to be redone for every component that grows this problem, and diverges from what `cldr:data:import` emits.
- **Deep-merge inside a custom I18n backend.** Fixes every key at once, but changes lookup semantics for the host application's own translations, not just CLDR's, and `I18n`'s first-match behaviour is the right one for leaves.
- **Special-case `stand_alone` → `format` in `Gregorian`.** This was my first read of the bug, and it is wrong: the `stand_alone` → `format` alias already works (it lives in `root/calendars.yml` and `I18n` resolves symbol links against `fallback_original_locale`). Hard-coding it would have masked the real problem while leaving `se-FI`, `ar-MA` and the `en-001` tree broken.

`weekday_names` / `month_names` now also raise `MissingCalendarDataError` when an entry is still missing after resolution. Previously `Worldwide::I18nExceptionHandler#degraded_translation` turned an unresolved structural key into a humanized copy of its last segment, so a missing weekday table came back as the **string** `"wide"` and `month_names` then died on `String#values` some distance from the cause. Degrading a leaf into readable text is reasonable; degrading a hash-valued CLDR node into a plausible-looking string is not.

- `Worldwide::Cldr#unresolved_node` reads a single locale's own contribution with `default: nil, fallback: false, resolve: false`. `default: nil` is what keeps a miss returning `nil` instead of going through the exception handler, and `resolve: false` is what returns a CLDR alias as a raw `Symbol` so it can be re-resolved against the requested locale.
- Alias resolution recurses. The `seen` set guards against two aliases pointing at each other; I did not find such a cycle in the current data, it is there so a future CLDR import cannot hang.
- `resolved_hash` is public because it is generally useful for structural CLDR keys, but nothing else in the gem needs it today — `Cldr::DateFormatPattern` only reads leaves. Happy to make it private if you'd rather not add API surface.
- Both methods now key their result off a canonical list (`sun..sat`, `1..12`) rather than the hash's insertion order.

Bug fix, no API change for callers that were already getting complete data. Two behaviour changes worth calling out:

- 104 locales start returning complete names. Anything snapshotting the truncated output will need updating.
- A structurally missing key now raises `Worldwide::Calendar::Gregorian::MissingCalendarDataError` instead of returning a string or raising `NoMethodError`. With this change `root` always supplies a complete set, so this should be unreachable with the shipped data.

### Testing

`bundle exec rake test` → 2256 tests, 0 failures. `bundle exec rubocop` → clean.

New coverage:

- `test/worldwide/calendar/gregorian_test.rb` asserts complete, non-empty weekday and month names for **all 805 known locales at all three widths**, plus explicit cases for the alias path (`en`, `nb`, which holds no calendar data of its own and inherits from `no`) and the partial-override path (`en-CA`, `en-GB`, `ar-MA`, `se-FI`).
- `test/worldwide/cldr_test.rb` covers `resolved_hash` directly, including that it returns `{}` for a missing key where `t` returns the degraded `"wide"`.

103 of the new per-locale assertions fail on `main` without the `lib/` change.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Security in case of vulnerabilities.
 
 ## [Unreleased]
+- Fix `Worldwide::Calendar::Gregorian.weekday_names` and `.month_names` returning partial results for 104 locales, including `en-CA`, `en-GB`, `en-IN` and every other `en-001` descendant. CLDR inheritance is item by item, but `I18n` returns the first node it finds in the fallback chain, so a locale that overrides a single entry (`en-CA` overrides only `Sept`) hid the rest of the entries it inherits: `month_names(locale: "en-CA", width: :abbreviated)` returned `["Sept"]` instead of all twelve names. Both methods now resolve names with CLDR's inheritance rules and raise `MissingCalendarDataError` instead of silently returning a degraded key. [#582](https://github.com/Shopify/worldwide/pull/582)
 
 ---
 
--- a/lib/worldwide/calendar/gregorian.rb
+++ b/lib/worldwide/calendar/gregorian.rb
@@ -3,26 +3,50 @@
 module Worldwide
   module Calendar
     class Gregorian
+      class MissingCalendarDataError < StandardError; end
+
       class << self
         # We intentionally don't support `short` weekday names, as we haven't found anyone asking for them.
         # People usually want the full weekday name, or the abbreviated version.
         VALID_WEEKDAY_WIDTHS = [:abbreviated, :narrow, :wide].freeze
+        WEEKDAYS = [:sun, :mon, :tue, :wed, :thu, :fri, :sat].freeze
         def weekday_names(width: :wide, locale: I18n.locale)
           raise ArgumentError, "Invalid width: #{width}" unless VALID_WEEKDAY_WIDTHS.include?(width)
 
-          Worldwide::Cldr.t("calendars.gregorian.days.stand_alone.#{width}", locale: locale)
+          complete_names("calendars.gregorian.days.stand_alone.#{width}", WEEKDAYS, locale)
         end
 
         VALID_MONTH_WIDTHS = [:abbreviated, :narrow, :wide].freeze
+        MONTHS = (1..12).to_a.freeze
         def month_names(width: :wide, locale: I18n.locale)
           raise ArgumentError, "Invalid width: #{width}" unless VALID_MONTH_WIDTHS.include?(width)
 
-          Worldwide::Cldr.t("calendars.gregorian.months.stand_alone.#{width}", locale: locale).values
+          complete_names("calendars.gregorian.months.stand_alone.#{width}", MONTHS, locale).values
         end
 
         def quarter(date, locale: I18n.locale)
           format_string = Worldwide::Cldr.t("calendars.gregorian.additional_formats.yQQQ", locale: locale)
           Worldwide::Cldr::DateFormatPattern.format(date, format_string, locale: locale)
+        end
+
+        private
+
+        # `stand_alone` is an alias to `format` in CLDR for most locales, and many locales
+        # override only a handful of entries, so these names have to be assembled with
+        # CLDR's inheritance rules rather than read out of a single locale's data.
+        #
+        # Anything still missing afterwards is a data bug. Raise rather than let it through:
+        # `Worldwide::Cldr.t` degrades a missing key into a humanized version of its last
+        # segment, so the alternative is silently handing back the string "wide".
+        def complete_names(key, expected_keys, locale)
+          names = Worldwide::Cldr.resolved_hash(key, locale: locale)
+          missing = expected_keys - names.keys
+
+          unless missing.empty?
+            raise MissingCalendarDataError, "Missing CLDR data for '#{key}' in locale '#{locale}': #{missing.join(", ")}"
+          end
+
+          expected_keys.to_h { |name_key| [name_key, names.fetch(name_key)] }
         end
       end
     end
--- a/lib/worldwide/cldr.rb
+++ b/lib/worldwide/cldr.rb
@@ -68,7 +68,63 @@ module Worldwide
         CONFIG
       end
 
+      # Look up a structural (hash-valued) CLDR key with CLDR's own inheritance rules.
+      #
+      # CLDR resolves locale data item by item: the effective data for a locale is the
+      # union of its own data with each of its ancestors', the more specific locale
+      # winning per item, and `<alias>` elements resolved against the locale originally
+      # requested (UTS #35 §4.1 "Multiple Inheritance" and §4.4 "Alias Elements").
+      #
+      # `I18n`'s fallback backend instead returns the first non-nil node it finds in the
+      # chain, which is correct for leaves but wrong for whole nodes: a locale that
+      # overrides a single entry hides every sibling entry it should inherit. `en-CA`
+      # only overrides September ("Sept" rather than "Sep"), so a plain `t` of the
+      # abbreviated month names returns a one-entry hash instead of all twelve.
+      #
+      # Leaf lookups should keep using `t`; only whole-node lookups need this.
+      def resolved_hash(key, locale: I18n.locale)
+        with_cldr do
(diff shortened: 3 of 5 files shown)


nightsurgex2:
Companion PR for the byte-identical copy in `shopify-i18n`: Shopify/shopify-i18n#2300

najibna (commented):
Right fix: I18n first-match was truncating CLDR item-by-item inheritance. Raising `MissingCalendarDataError` instead of degrading to `"wide"` is correct, and the en-CA / nb / se-FI cases are the ones that matter.

PR #592, opened 2026-09-09
Add `:date_time_current_year_with_seconds` date/time format

SimonSiju:
## What

Adds `:date_time_current_year_with_seconds` — `:date_time_current_year` (a compact, year-omitted "MMM d + time" timestamp) with the time upgraded to **second precision**.

It exists for cross-system log/event correlation, where the exact second matters. Like the base format, **each locale owns the field order, separators, and 12h/24h clock** — nothing is forced:

| locale | renders |
|---|---|
| en | `Dec 26, 1:23:45 pm` (12-hour) |
| fr | `26 déc. 13:23:45` (24-hour, day-first) |
| vi | `13:23:45, 26 Thg 12` (time-first) |
| da | `26. dec. 13.23.45` (`.` separators) |
| ja | `12月26日 13:23:45` |
| ko | `12월 26일 오후 1:23:45` |

## How

- One entry in `DATETIME_FORMAT_MAP`: the locale's `datetime.short` combiner + `MMMd` + `time.medium` (the seconds-bearing, locale-clock time format).
- Regenerated `data/other/generated/*.yml` (805 locales) + `formats.md` — reproducible via the CLDR rake task (`bin/test-cldr-rake-tasks` re-runs it and fails on any diff).
- Test: representative locales + an all-locales guard asserting each locale's pattern carries a seconds field.

## Notes (review feedback addressed)

- **Named for what it is** — it's `date_time_current_year` + seconds, not `date_at_time` + seconds; the name reflects that so it doesn't need a breaking rename later.
- **Uses each locale's own clock** (`time.medium`) rather than forcing 24h everywhere.
- **Year is omitted**, matching `date_time_current_year`. The consuming surface (Dev Dashboard app-logs) shows recent logs and pins the exact instant via its local + UTC hover tooltip.
- Only 3 hand-written files (generator, test, CHANGELOG); everything under `data/other/generated/` is generated.

Consuming change: Dev Dashboard app-logs timestamps (shop/issues-develop#23821).


najibna (commented):
Reusing `date_time_current_year` with `time.medium` is the right approach, and the all-locales `%S` assertion is a good guard.

Nit: the changelog entry is longer than the rest of this file. The first sentence is enough.

PR #593, opened 2026-09-10
Add the 81 Türkiye province zones, staged behind ignore_provinces

DominiqueFlaaa:
## What

81 province (il) zones for Türkiye in `data/regions/TR.yml`, staged: `ignore_provinces: true`, `hide_provinces_from_addresses: true`, no `{province}` in the formats. Step 1 of shop/issues-addresses#1131.

## Data decisions, each from precedent

- **Codes**: `code` = `iso_code` = `TR-01`..`TR-81`, the ISO 3166-2:TR subdivision codes (plaka numbers). Validate against the official list: https://www.iso.org/obp/ui/#iso:code:3166:TR. Every zone set added since 2020 uses prefixed ISO codes (PH-ABR, KW-AH, CR-A, SV-AH, UY-AR, VE-Z); JP-01 is the numeric precedent. Bare codes (CA, US, GB) are all legacy.
- **Names**: CLDR English subdivision names (`tr01`..`tr81`), the KW convention. Translations already ship in `data/cldr/locales/*/subdivisions.yml`, no translation work.
- **Order**: alphabetical by `ActiveSupport::Inflector.transliterate(name.downcase)`, per the consistency test. Not plaka order: Aksaray (68) sorts after Ağrı (04).
- **zip_prefixes**: each zone's plaka number. The first two digits of a Turkish postal code are the province number, so provinces are inferable from postal codes (55.2% of stored TR addresses, measured on the issue). Prefix `99` belongs to Mersin (TR-33): Northern Cyprus is addressed via Türkiye as "Mersin 10" with the 99xxx range, and `Zip.find_country(country_code: :CY, zip: "99628")` must keep returning TR.
- **example_address**: gains `province_code: TR-34` (zip 34122 is İstanbul), required by the consistency tests once zones exist. The city drops the province suffix (`Fatih/İstanbul` → `Fatih`): with zones defined, the il belongs in the province field, not the city.

## Tax: `tax: 0.0` per zone, no `tax_name`

The `tax:` key is mandatory: `region_yml_consistency_test.rb` ("all countries and provinces have tax information") asserts every province's `tax_rate` is a Float in 0..1, and the loader passes the YAML value through with no zone-level default (`regions_loader.rb`, `tax_rate: zone["tax"]`), so omitting the key fails the suite.

The value 0.0 is verified: KDV is levied nationally under Law 3065, with rates (20%/10%/1%) set by national legislation and Presidential decision. No province-level VAT or regional surtax exists across the 81 il ([Grant Thornton indirect tax guide, Türkiye](https://www.grantthornton.global/en/insights/indirect-tax-guide/indirect-tax---turkey/), [Kreston VAT guide, Türkiye](https://www.kreston.com/vat-guide/turkey/)). The national rate stays at the country level (`tax: 0.18` in this file); zones contribute nothing on top, the same shape as PH, KW, CR, SV, UY, VE.

`tax_name` is omitted: the consistency test only requires a tax name when the zone rate is non-zero.

## What does not change

The staged flag pair plus no `{province}` in the formats is a legal state for `region_data_consistency_test`. `has_provinces?` stays false, address forms and `AddressValidator` behavior are untouched until go-live removes the flags.

## Verification

- Full suite: 1445 tests, 0 failures.
- `region(code: "TR").zone(zip: "34122")` → TR-34, `zone(zip: "01960")` → TR-01, `zone(zip: "99628")` → TR-33.
- `zone(code: "TR-04")` and `zone(code: "04")` both resolve (lookup registers both forms).

## Sequencing

After this releases: country_db zone stubs (with Shopify/country_db#2061 already merged there), then core → pay → billing → Atlas last, so the shipping picker cannot show regions the backend rejects.

### DIFF
--- a/data/regions/TR.yml
+++ b/data/regions/TR.yml
@@ -6,6 +6,8 @@ currency: TRY
 unit_system: metric
 tax_name: KDV
 tax_inclusive: true
+ignore_provinces: true
+hide_provinces_from_addresses: true
 group: European Countries
 group_name: Europe
 zip_example: '01960'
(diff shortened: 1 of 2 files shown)


DominiqueFlaaa:
## Validation against real addresses

We verified the prefix-to-province mapping against an internal sample of 10,000 real Turkish addresses. 3,719 of them carried a valid 5-digit postal code plus an independently identified il.

| Outcome | Rows | Share |
|---|---:|---:|
| Identified il agrees with this mapping | 3,507 | 94.3% |
| Identified value was not an il (mahalle or ilçe) | 154 | 4.1% |
| Identified value was a `TR-xx` code naming the same il | 2 | 0.1% |
| Genuine il-vs-postal-code conflicts | 56 | 1.5% |

The 1.5% conflicts are address data errors, not mapping errors: in those rows the stated city and the postal code point at different provinces (placeholder codes, or a postal code that belongs to a neighbouring il). A mapping error would show as one consistently repeated disagreement pair; no pair occurred more than twice and the directions are random.

The non-il rows also support the mapping where checkable: district-level values resolved to the province the postal code predicts (e.g. an ilçe of Antalya on an Antalya-prefixed code).

No postal codes with prefixes 82-98 appeared in the sample, and unknown prefixes were almost entirely `000xx` placeholders.

data/regions/TR.yml
```diff
@@ -34,7 +36,497 @@ languages:
...
+  - '06'
+- name: Antalya
+  code: TR-07
+  iso_code: TR-07
+  tax: 0.0
+  zip_prefixes:
+  - '07'
+- name: Ardahan
+  code: TR-75
+  iso_code: TR-75
```
toneymathews: it's not odd that the order of the code breaks with TR-75 coming between 07 and 08 right? Probably not, just wanted to be sure

toneymathews (approved):
lgtm!

PR #596, opened 2026-09-16
Reject malformed currency codes before caching

o-benz:
## Context

PR 1/? of https://github.com/shop/issues/issues/84888. Malformed currency identifiers pass substring validation and remain in the process cache. [PR276](https://github.com/Shopify/worldwide/pull/276) left this cache unresolved.

## Approach

Require the entire normalized code to be three ASCII letters or three ASCII digits before it can be cached, preserving existing normalization and custom and historical currencies. The same bounded reproduction adds eight malformed entries on main and none with this fix, and regression tests cover both the public facade and direct constructor. Local Ruby matrix and TypeScript checks passed. GitHub CI requires a PR or a push to main, so remote CI has not run for this branch.

### Checklist


SeriouslyAwesome (approved):
Makes sense to me 👍

PR #598, opened 2026-09-17
Add an ISO 3166-1 assignment predicate

chrisdakin:
Add Region#iso_3166_code_assigned? to identify regions with currently assigned ISO 3166-1 alpha-2 codes.

Check the resolved region's iso_code against a frozen set of 249 assigned codes.

Reference: https://www.iso.org/iso-3166-country-codes.html


rochlefebvre (approved):

alaar (approved):

najibna (commented):
The predicate is clear, and the tests cover assigned codes, reserved/user-assigned codes, subdivisions, and lookup aliases.

One note: the YAML snapshot cites Wikipedia. A pointer to the ISO 3166-1 Maintenance Agency list would make the source of the 249 codes easier to re-check later.

PR #605, opened 2026-09-20
Treat address replacement values as literal data

### LINKED ISSUE #549: Address values can be interpreted as format tokens
## Problem

Address values containing Worldwide format tokens can be interpreted as later placeholders because fields are substituted sequentially.

```ruby
address = Worldwide::Address.new(
  first_name: "{lastName}",
  last_name: "Smith",
  address1: "{phone}",
  country_code: "US",
  phone: "secret",
)

address.format(excluded_fields: [:phone])
```

Worldwide 1.25.6 formats the recipient as `Smith Smith` instead of preserving the literal first name `{lastName}`. It can also expand the literal address value `{phone}` to the phone data even though the phone field was excluded.

Additional lines are processed through the same substitution path and have the same behavior.

## Cause

`Address#fill_in_fields` substitutes each field in sequence. A replacement value inserted early can contain a token that is matched during a later field substitution.

Using block replacement, as proposed in https://github.com/Shopify/worldwide/pull/533, fixes backslash interpretation but does not prevent a token inserted for one field from being processed during a later field's substitution.

## Expected behavior

Address values and additional line values should be treated as literal data. They should not be interpreted as format syntax, and excluded field values should not be reintroduced through user supplied tokens.

This was found while migrating `ShopifyI18n::AddressFormatter` to delegate directly to Worldwide. The downstream adapter should not need to escape Worldwide's internal format syntax.

najibna:
## Description

Fixes #549

Address format tokens are now replaced in a single pass. User-supplied values such as `{lastName}` or `{phone}` are treated as literal data and cannot be interpreted as later format tokens, including when those fields are excluded.

## How has this been tested?

- `bundle exec ruby -Itest test/worldwide/address_formatting_test.rb` — 43 tests, 0 failures

## Checklist

- [x] Tests updated
- [x] CHANGELOG left unchanged per CONTRIBUTING.md


najibna:
I have signed the CLA!

